PR #13, opened 2019-04-10
removed guiding config state field and added a boolean optional field…

jnation3406:
… to account for optional guiding of any mode


jnation3406:
Since we currently only have one 'guide' mode, I anticipate keeping the names OFF and ON in configdb for our current cameras. Future cameras with multiple modes can name them more descriptively. 

So things map to the new fields: 
ag_mode ON = mode ON, optional False
ag_mode OPTIONAL = mode ON optional True
ag_mode OFF = mode OFF, optional False

observation_portal/requestgroups/serializers.py
```diff
@@ -168,26 +168,21 @@ def validate(self, data):
...
                     and guiding_config['mode'].lower() not in [gm['code'].lower() for gm in modes['guiding']['modes']]):
-                raise serializers.ValidationError(_("guiding mode {} is not available for instrument type {}"
-                                                    .format(guiding_config['mode'], data['instrument_type'])))
+            raise serializers.ValidationError(_("Guiding mode {} is not available for instrument type {}"
+                                                .format(guiding_config['mode'], data['instrument_type'])))
+
+        if configdb.is_spectrograph(data['instrument_type']):
+            if 'optional' in guiding_config and guiding_config['optional'] and data['type'] != 'ARC':
+                raise serializers.ValidationError(_("Guiding must not be optional on spectrograph instruments and non-ARC observations."))
+            guiding_config['optional'] = False
```
eheinrich: I think this behavior is different from before- If `guiding_config['optional']` is `True`, and the observation is of type ARC, optional will be set to False. Is this what we want?

jnation3406: good call, I guess I should move the ARC check to the outer if statement

PR #14, opened 2019-04-16
Vue request details

jnation3406:


jnation3406:
I think the request details and list pages are mostly good (except the jquery import issue on the request list page, which just affects the pagination buttons right now). I have not yet moved them to all-vue

PR #15, opened 2019-04-22
Request shim and time allocation command

jnation3406:


jnation3406:
I've tested the validate, submission, cadence, and max_ipp stuff by hand and with scripts. Doesn't seem worth tons of unit tests for a temporary shim. The time allocation stuff is still a work in progress.

observation_portal/observations/management/commands/time_accounting.py
```diff
@@ -0,0 +1,83 @@
...
+
+        semester = Semester.objects.get(id=options['semester'])
+        for proposal in proposals:
+            for instrument_type in instrument_types:
+                attempted_time = {
+                    'NORMAL': 0,
+                    'RAPID_RESPONSE': 0,
+                    'TIME_CRITICAL': 0
+                }
+                observations = Observation.objects.filter(end__gt=semester.start, start__lt=semester.end,
```
eheinrich: I think function assumes that each request will only be for a single type of instrument, is that right? I would just add a note saying that.

observation_portal/userrequests/conversion.py
```diff
@@ -0,0 +1,249 @@
...
+
+def convert_userrequests_to_requestgroups(userrequests):
+    if isinstance(userrequests, list):
+        return [convert_userrequest_to_requestgroup(ur) for ur in userrequests]
+    else:
+        return convert_userrequest_to_requestgroup(userrequests)
+
+
+def convert_userrequest_to_requestgroup(userrequest):
+    requestgroup = userrequest
```
eheinrich: It doesn't look like it matters in here, but any change made to `requestgroup` will also be made to `userrequest` since they are the same underlying object.

observation_portal/userrequests/conversion.py
```diff
@@ -0,0 +1,249 @@
...
+
+def convert_requestgroups_to_userrequests(requestgroups):
+    if isinstance(requestgroups, list):
+        return [convert_requestgroup_to_userrequest(rg) for rg in requestgroups]
+    else:
+        return convert_requestgroup_to_userrequest(requestgroups)
+
+
+def convert_requestgroup_to_userrequest(requestgroup):
+    userrequest = requestgroup
```
eheinrich: Same comment

PR #19, opened 2019-05-09
Increase rapid response lifetime to 24 hours

eheinrich:
Rapid response observation window end times were limited to be within the next 6 hours in the future, this PR increases that limit to be within 24 hours in the future.

### DIFF
--- a/static/js/components/util/customdatetime.vue
+++ b/static/js/components/util/customdatetime.vue
@@ -22,7 +22,7 @@
         </sup>
       </template>
       <VueCtkDateTimePicker 
-        v-model="theValue"
+        v-model="value"
         label=""
         :format="datetimeFormat"
         :formatted="datetimeFormat"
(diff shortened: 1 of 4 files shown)


static/js/components/util/customdatetime.vue
```diff
@@ -70,8 +70,7 @@
     data: function() {
       return {
         tooltipConfig: tooltipConfig,
-        datetimeFormat: datetimeFormat,
-        theValue: this.value
```
jnation3406: Is value an internal field so you don't need this?

eheinrich: Yep, it's a prop. If you're wondering how there is a `value` prop being passed into the `customdatetime` component, it's from using `v-model`: https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components

PR #21, opened 2019-05-18
changed target types to match api

jnation3406:


jnation3406:
I'm going to merge this in because i need to test it in prod with Mark's code. We can fix any issues in master

PR #22, opened 2019-05-20
Feature/documentation

eheinrich:
Besides updating some help text, links, and icons, this PR also:
- Implements generic rotator modes
- Updates the `/observations/` page to exclude cancelled observations if the only query parameter is for pagination
- Updates the `/tools/` page to retrieve the instrument types used in the selects for the pressure and contention plots from the API
- Display an error message on the compose page when the user is not a member of an active proposal


static/js/components/requestgroup.vue
```diff
@@ -245,6 +261,13 @@
       }
     },
     watch: {
+      proposalOptions: function() {
+        if (this.hasRetrievedProposals && this.proposalOptions.length < 2) {
```
jnation3406: why is this < 2?

eheinrich: Because the first item in the proposals options list is always an empty option. I'll add a comment since that's not clear.

PR #24, opened 2019-06-06
display proper fields for direct type observation on request details …

jnation3406:
…page


observation_portal/requestgroups/models.py
```diff
@@ -193,8 +194,15 @@ def as_dict(self, for_observation=False):
         ret_dict['duration'] = self.duration
         ret_dict['configurations'] = [c.as_dict() for c in self.configurations.all()]
         if not for_observation:
-            ret_dict['location'] = self.location.as_dict()
-            ret_dict['windows'] = [w.as_dict() for w in self.windows.all()]
+            if self.request_group.observation_type == RequestGroup.DIRECT:
+                Observation = apps.get_model('observations', 'Observation')
```
eheinrich: Why did you have to get the observation like this?

jnation3406: circular reference (Observation models reference the Request models)

eheinrich: Would it be better to get the observation by doing something like `self.observation_set.first()`?

jnation3406: Yeah you are right. For some reason I didn't think this would work but it seems to be fine.

PR #29, opened 2019-06-10
Fixes during testing

eheinrich:
There are a few different things going on in this PR, sorry for the bulk! Mainly:
- Fix for determining the email address for the notification to the PI when a proposal is ported (this was a semester change error)
- Error handling when setting the global time limit for CIs on a proposal in case the value passed in can not be converted to a float (we also got an error email for this around the semester change time)
- In the configdb file I've made a change to exclude instruments that are disabled.
- In the requestgroup serializer, I've refactored the validation for the various modes to decrease some duplicated code.
- In general, I did some refactoring for failed pep8 validation, switching to python3.6 string formatting, etc.


observation_portal/requestgroups/serializers.py
```diff
@@ -25,10 +26,83 @@
...
 
 
+class ModeValidator:
+    """Class used to validate modes of different types"""
+    def __init__(self, mode_type, instrument_type, default_modes, modes):
+        self._mode_type = mode_type.lower()
+        self._instrument_type = instrument_type
+        self._default_modes = default_modes
+        self._modes = modes
+        self._mode_key = 'rotator_mode' if self._mode_type == 'rotator' else 'mode'
```
jnation3406: why is this different for rotator?

observation_portal/sciapplications/models.py
```diff
@@ -210,7 +210,11 @@ def send_approved_notification(self):
                 'semester_already_started': self.call.semester.start < timezone.now(),
             }
         )
-        send_mail.send(subject, message, 'portal@lco.global', [self.proposal.pi.email])
+        # Find the email to send the notification to. The proposal will have been created at this point, but the pi on
+        # the proposal might not be set yet if the pi has not a registered an account. In that case, use the email
+        # as set on the science application.
+        pi_email = self.proposal.pi.email if self.proposal.pi else self.pi
+        send_mail.send(subject, message, 'portal@lco.global', [pi_email])
```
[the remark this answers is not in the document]
eheinrich: Yeah, good idea. I'll add something in.

observation_portal/requestgroups/serializers.py
eheinrich: Because in the configuration dict (which could be the instrument_config or guiding_config... whatever is passed in) the rotator mode is the set in the `rotator_mode` field, and all the other modes are just set in the `mode` field in their respective configuration dicts.

PR #30, opened 2019-06-10
Codify fixes

jnation3406:


static/js/request_row.js
```diff
@@ -1,7 +1,6 @@
 import $ from 'jquery';
 import moment from 'moment';
 import {datetimeFormat} from './utils.js';
-import tooltip from 'bootstrap';
```
jnation3406: I don't know why this was added in the observation-portal, but I couldn't see any difference after removing it and codify complains it's unused.

static/js/utils.js
```diff
@@ -94,21 +93,20 @@ function decimalRaToSexigesimal(deg){
...
+  let ds = 1;
+  let dec = deg;
   if(deg < 0){
     ds = -1;
-    var dec = Math.abs(deg);
-  } else {
-    var dec = deg;
+    dec = Math.abs(deg);
   }
-  var deg = Math.floor(dec)
```
jnation3406: We had overwritten the deg variable that was passed in before... so im not sure if this equation is correct. We should double check this.

static/js/request_row.js
eheinrich: Probably just a leftover from before that wasn't removed

static/js/utils.js
eheinrich: This looks fine to me

static/js/components/pressure.vue
```diff
@@ -137,6 +137,21 @@
...
+          }
+        }
+        return false;
+      }
+    },
+    created: function() {
+      this.fetchData();
+    },
+    methods: {
       toSiteNightData: function() {
```
eheinrich: If `toSiteNightData` is now a method, doesn't that break where it is used?

jnation3406: ah I guess I should add () where we call the method...

PR #32, opened 2019-06-13
switch back certain configdb methods to allowing DISABLED instruments

jnation3406:


observation_portal/common/configdb.py
```diff
@@ -477,14 +480,14 @@ def get_active_instrument_types(self, location: dict) -> set:
         return instrument_types
 
     def get_guider_for_instrument_name(self, instrument_name):
```
eheinrich: Should this method and the next (`is_valid_guider_for_instrument_name`) exclude disabled instruments?

jnation3406: I guess so

PR #34, opened 2019-06-18
name Observation Portal consistently

phycodurus:
I'm embarrassed to be so punctilious about naming, but I've changed all the occurrences of

- "observing portal" to "observation portal"
- "observing request" to "observation request"

"observing windows" were not changed.


phycodurus:
That's a good idea. The document "Getting Started on the LCO Global Telescope Network" has the same issue and  I'm happy to touch base with Nikolaus about it.

PR #36, opened 2019-06-20
Fix/compose page issues

eheinrich:
Things in this PR:
- Added text to the compose page buttons, and removed their tooltips
- Always set the acquire modes in the compose page, even for imagers. If a camera does not have any acquire modes, set to OFF.
- Comments from Nikolaus- updated some help text, fixed a bug in how configdb compiles optical elements.
- Worked through how bootstrap and javascript libraries are loaded in the django templates.
- In the instrumentconfig, update the binning even when the readout mode code does not change to account for the case where two different modes have the same code, but different binnings.
- Fix target lookup service, a leftover from the target type change.
- Move ajax setup out of global entrypoint and into specific entry points where the setup is used. The jquery docs warned against the way we were doing it, and now the global bundle is a bit smaller.


observation_portal/common/configdb.py
```diff
@@ -334,13 +334,15 @@ def get_optical_elements(self, instrument_type: str) -> dict:
...
+        optical_elements = defaultdict(list)
+        optical_elements_tracker = defaultdict(set)
         for instrument in self.get_instruments(exclude_states=['DISABLED', ]):
             if instrument_type.upper() == instrument['science_camera']['camera_type']['code'].upper():
                 for optical_element_group in instrument['science_camera']['optical_element_groups']:
-                    optical_elements[optical_element_group['type']] = []
                     for element in optical_element_group['optical_elements']:
-                        optical_elements[optical_element_group['type']].append(element)
+                        if element['code'] not in optical_elements_tracker[optical_element_group['type']]:
+                            optical_elements_tracker[optical_element_group['type']].add(element['code'])
```
jnation3406: This might work by just returning the dict of sets rather than the lists. The serializer usage just looks like it is using `in` so it should be fine, so its just the stuff passed to the instrument info view, which might work out fine too with a set vs list.

static/js/components/sidenav.vue
```diff
@@ -320,4 +323,10 @@
   .request li ~ ul.nav {
     display: none;
   }
+  hr {
```
jnation3406: what is this for?

static/js/components/util/panel.vue
```diff
@@ -8,18 +8,18 @@
               <b-form-row>
                 <b-col class="text-left">
                   <i 
-                    class="align-middle mx-2" 
+                    class="align-middle fa-lg mx-2"
```
jnation3406: the fa-lg works without having a font awesome icon?

static/js/components/window.vue
```diff
@@ -68,15 +68,15 @@
...
+              desc="Decimal hours"
               :errors="errors.period"               
               @input="update"
             />
             <customfield v-show="cadence === 'simple'"
               v-model="jitter" 
               label="Jitter" 
               field="jitter" 
-              desc="Acceptable deviation (before or after) from strict period."
+              desc="Acceptable deviation in decimal hours (before or after) from strict period."
```
jnation3406: I would phrase this "Acceptable deviation from strict period (before or after) in decimal hours.

templates/base.html
```diff
@@ -1,5 +1,6 @@
 <!doctype html>
 {% load static bootstrap4 %}
+{% bootstrap_javascript jquery='full' %}
```
jnation3406: so this definitely works just having it in the base? I thought we had tried that and needed it in the child pages or the javascript controls didn't work...

observation_portal/common/configdb.py
eheinrich: The thing this is compiling looks something like `{'filters': [{'code': 'V', 'name': 'Bessell-V'}, ... ]}`. Each container for the various optical element types contains a collection of dicts, and dicts are not hashable so they cannot go into a set. The instrument info view is used by the compose page, which needs more than just the code of each optical element.

static/js/components/sidenav.vue
eheinrich: There is a subtle horizontal line under the buttons in the sidenav.

static/js/components/util/panel.vue
eheinrich: Oh, this does actually get a font awesome icon. On the next line there is `:class="icon"` - this is binding the icon class that is passed into the component as a prop.

templates/base.html
eheinrich: Earlier it was working on some pages, for some sections, but not all the time. There was actually some javascript that was being loaded in by a vue entrypoint in one of the sub templates that was messing up the javascript that had been loaded like this... another change that I made in this PR was moving some of the extra_javascript stuff out of the block content sections in various templates. After sorting through that stuff, things seemed to be working, but bootstrap and the extra javascript libraries were being loaded multiple times, once each for each time this bootstrap_javascript tag was included. It does look to work like this.

observation_portal/common/configdb.py
jnation3406: Oh yeah, I didn't realize that, this seems fine then

static/js/components/util/panel.vue
jnation3406: ahhh

static/js/components/window.vue
eheinrich: Done

PR #37, opened 2019-06-24
Feature/set versions

eheinrich:


Dockerfile
```diff
@@ -26,7 +26,7 @@ CMD gunicorn observation_portal.wsgi -c observation_portal/config.py
 COPY requirements.txt .
 RUN apt-get update \
   && apt-get install -y gfortran libcairo2 libpango-1.0-0 libpangocairo-1.0-0 \
-  && pip install 'numpy<1.16' \
```
eheinrich: I don't remember why we put this restriction on numpy, do you?

PR #46, opened 2019-07-02
Science support fixes

eheinrich:
Fixes:
- Optical element not populated when loading an observation request. This was because the dropdown values were case sensitive. I've just lowercased all optical element codes in the dropdowns, I think this should be ok but let me know what you think.
- Defocus not populated when loading observation request.
- Fill window calculation bug
- Allow eccentricity values >= 1
- Add filter by requestgroup name to front page
- Fix url for observations associated with a request


jnation3406:
sorry I broke a test in one of my changes for ASAS last night, I fixed it in my latest pull request with priority

static/js/request_row.js
```diff
@@ -36,7 +36,7 @@ $(document).ready(function() {
   $('.pending-details').each(function() {
     let that = $(this);
     let requestId = $(this).data('request');
-    $.getJSON('/api/requests/' + requestId + '/?exclude_canceled=true', function(data) {
+    $.getJSON('/api/requests/' + requestId + '/observations/?exclude_canceled=true', function(data) {
```
jnation3406: I wonder how this got in here... this should have never worked with that URL

eheinrich: Yeah, I guess it probably just slipped by unnoticed :woman_shrugging:

eheinrich:
Ok, I'll wait until the other branch is merged in to master, then I'll get the branch passing and merge it in

PR #49, opened 2019-07-11
Feature/more science support

eheinrich:
Fixes:
- Change from "Title" to "Name" on compose page requestgroup for consistency
- Paginate and filter CoIs when there are lots of them
- Fix a race condition where the first time that a user logs into the archive by going to a request or requestgroup page, the token is not stored yet and then would have to refresh the page in order to see the thumbnails.

### DIFF
--- a/observation_portal/proposals/templates/proposals/proposal_detail.html
+++ b/observation_portal/proposals/templates/proposals/proposal_detail.html
@@ -16,7 +16,18 @@
         </h4>
         {% if user == object.pi %}
         <h4>Co Investigators <small><i id="toggleci" class="fa fa-eye"></i></small></h4>
-        <table class="table" id="cilist">
+        <div id="cilist">
+          {% if members_are_paginated %}
+          <form method="get" role="form" action="">
+            <div class="form-row">
+              <div class="col mb-0 mt-auto">{% bootstrap_field members_filter.form.first_name form_group_class="form-group mb-0 mt-auto" %}</div>
+              <div class="col mb-0 mt-auto">{% bootstrap_field members_filter.form.last_name form_group_class="form-group mb-0 mt-auto" %}</div>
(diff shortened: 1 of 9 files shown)


observation_portal/proposals/templates/proposals/proposal_detail.html
```diff
@@ -16,9 +16,22 @@ <h4>
...
         <h4>Co Investigators <small><i id="toggleci" class="fa fa-eye"></i></small></h4>
-        <table class="table" id="cilist">
+        <div id="cilist">
+          {% if members_are_paginated %}
+          <form method="get" role="form" action="">
+            <div class="form-row">
+              <div class="col mb-0 mt-auto">{% bootstrap_field members_filter.form.first_name form_group_class="form-group mb-0 mt-auto" %}</div>
+              <div class="col mb-0 mt-auto">{% bootstrap_field members_filter.form.last_name form_group_class="form-group mb-0 mt-auto" %}</div>
+              <div class="col mb-0 mt-auto">{% bootstrap_field members_filter.form.username form_group_class="form-group mb-0 mt-auto" %}</div>
+              <div class="col mb-0 mt-auto">{% bootstrap_field members_filter.form.email form_group_class="form-group mb-0 mt-auto" %}</div>
```
jnation3406: could probably just go with a single generic "filter" field that checks for that text in name, email, username. But this works too.

eheinrich: Oh, do you have an example of how to do that? I was trying to do something like that but couldn't really figure out in a reasonable amount of effort a good way so just did this. I was able to easily get a single input to filter against many fields, but it was an AND or OR across those fields, and I didn't think that looping through each object in the queryset, and comparing against concatenated fields was a good idea. Maybe I didn't come across the right docs.

PR #50, opened 2019-07-17
Remove "ObservationPortalRequestLog" log messages

irasnyd:
This application logs a bunch of lines which say
"ObservationPortalRequestLog" (and nothing else) over and over again.
This simply fills up the logging system's disk, without providing any
additional useful information (as far as I can tell).

Here is an example query showing all instances of this log message since
the current containers were started:

```
$ kubectl logs -l app.kubernetes.io/instance=observation-portal --all-containers --timestamps | grep ObservationPortalRequestLog
2019-07-17T22:18:15.786459777Z ObservationPortalRequestLog
2019-07-17T22:18:15.896831391Z ObservationPortalRequestLog
2019-07-17T22:18:16.267207083Z ObservationPortalRequestLog
2019-07-17T22:18:16.321665022Z ObservationPortalRequestLog
2019-07-17T22:18:17.158035736Z ObservationPortalRequestLog
2019-07-17T22:18:11.61618424Z ObservationPortalRequestLog
2019-07-17T22:18:13.392393172Z ObservationPortalRequestLog
2019-07-17T22:18:14.553258432Z ObservationPortalRequestLog
2019-07-17T22:18:16.568681714Z ObservationPortalRequestLog
2019-07-17T22:18:17.126812199Z ObservationPortalRequestLog
2019-07-17T22:18:15.128505218Z ObservationPortalRequestLog
2019-07-17T22:18:15.165644765Z ObservationPortalRequestLog
2019-07-17T22:18:16.819880902Z ObservationPortalRequestLog
2019-07-17T22:18:18.42285911Z ObservationPortalRequestLog
2019-07-17T22:18:19.03645736Z ObservationPortalRequestLog
2019-07-17T22:18:14.71166681Z ObservationPortalRequestLog
2019-07-17T22:18:17.651439984Z ObservationPortalRequestLog
2019-07-17T22:18:13.03176698Z ObservationPortalRequestLog
2019-07-17T22:18:14.256193229Z ObservationPortalRequestLog
2019-07-17T22:18:14.477296884Z ObservationPortalRequestLog
```

As far as I can tell, these logs aren't adding any value to the application. Since I couldn't figure out what they were for, I simply removed the code. This isn't honestly intended as a real fix, but I couldn't figure out what value these logs were supposed to be adding, either.

### DIFF
--- a/observation_portal/common/middleware.py
+++ b/observation_portal/common/middleware.py
@@ -6,46 +6,6 @@ from django.shortcuts import redirect
 from observation_portal.accounts.models import Profile
 
 
-class RequestLogMiddleware(object):
-    def __init__(self, get_response):
-        self.get_response = get_response
-        self.logger = logging.getLogger('observation_portal_request')
-
-    def __call__(self, request):
-        response = self.get_response(request)
-
-        try:
-            username = request.user.username
-            simple_interface = request.user.profile.simple_interface
-        except (AttributeError, Profile.DoesNotExist):
-            simple_interface = False
-            username = 'anonymous'
-
-        forwarded_ip = request.META.get('HTTP_X_FORWARDED_FOR')
-        if forwarded_ip:
-            ip_address = forwarded_ip.split(',')[0].strip()
-        else:
-            ip_address = request.META.get('REMOTE_ADDR')
-
-        tags = {
-            'ip_address': ip_address,
-            'uri': request.path,
-            'status': response.status_code,
-            'method': request.method,
-            'user': username,
-            'simple_interface': simple_interface,
-        }
-
-        if response.status_code >= 400:
-            level = logging.WARN
-        else:
-            level = logging.INFO
-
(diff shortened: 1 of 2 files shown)


irasnyd:
> This used to add some useful information to valhalla logs (see the fields in the tags dict in the RequestLogMiddleware, like username), but yeah you're right, they aren't useful anymore.

Yep. It wasn't obvious to me why the tags thing isn't working, but the proof is in the log messages. The tags don't show up in the raw unprocessed output (`kubectl logs ...`).

> I haven't looked into it yet, but I assume that fluentd doesn't parse those tags into fields as happened in old rancher. It's on my list to get this additional info into the logs again, but this is fine for now.

The fluentd parser is 100% compatible with the logstash parser (I have tests to prove it!), plus I made fluentd understand a bunch of our other common log formats, not just the "lco format". :smiley:

If you wish to examine the code, here it is:
https://github.com/LCOGT/elasticsearch-ingest-node-pipeline/blob/master/kubernetes-fluentd/pipeline.yaml

Thanks for looking at this!

PR #52, opened 2019-07-18
Set idle in transaction timeout

eheinrich:
I read through the [blog post](https://medium.com/squad-engineering/configure-postgres-statement-timeouts-from-within-django-6ce4cd33678a) describing how to set the timeout with a signal, but doing it that way did not feel quite right since I didn't find a clean way to reliably register that signal once (usually you register signals in each app, depending which one it goes with, but this would have needed to be registered once for the entire project on startup.)

After deciding that, I did a bit more digging and found that the parameter can be set in the settings file: https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-OPTIONS, and the backend options https://www.postgresql.org/docs/9.6/libpq-connect.html#LIBPQ-CONNECT-OPTIONS

### DIFF
--- a/observation_portal/settings.py
+++ b/observation_portal/settings.py
@@ -103,6 +103,9 @@ WSGI_APPLICATION = 'observation_portal.wsgi.application'
 # Database
 # https://docs.djangoproject.com/en/2.1/ref/settings/#databases
 
+# https://www.postgresql.org/docs/9.6/runtime-config-client.html#GUC-IDLE-IN-TRANSACTION-SESSION-TIMEOUT
+PORTAL_IDLE_IN_TRANSACTION_TIMEOUT = 60 * 60 * 1000  # 1 hour
+
 DATABASES = {
    'default': {
        'ENGINE': os.getenv('DB_ENGINE', 'django.db.backends.postgresql'),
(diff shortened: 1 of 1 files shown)


observation_portal/settings.py
```diff
@@ -103,14 +103,20 @@
...
    'default': {
        'ENGINE': os.getenv('DB_ENGINE', 'django.db.backends.postgresql'),
        'NAME': os.getenv('DB_NAME', 'observation_portal'),
        'USER': os.getenv('DB_USER', 'postgres'),
        'PASSWORD': os.getenv('DB_PASSWORD', ''),
        'HOST': os.getenv('DB_HOST', '127.0.0.1'),
-       'PORT': os.getenv('DB_PORT', '5432')
+       'PORT': os.getenv('DB_PORT', '5432'),
+       'OPTIONS': {
+           'options': f'-c idle_in_transaction_session_timeout={PORTAL_IDLE_IN_TRANSACTION_TIMEOUT}'
```
irasnyd: Which code picks this setting up and applies it to each database connection? I couldn't find any documentation that says Django does this for the PostgreSQL backend. The only examples I was able to find online showed the need to manually hook up the "database connected" signal and run the `SET SESSION ...;` SQL command ourselves in the code.

irasnyd: @eheinrich and I talked. She's tested this by setting a very short timeout, and then running a transaction which is idle for longer than the configured timeout, and seeing the expected behavior.

PR #53, opened 2019-07-18
added task for deleting old observations

jnation3406:


observation_portal/observations/models.py
```diff
@@ -72,6 +74,16 @@ def cancel(observations):
 
         return deleted_observations.get('observations.Observation', 0) + canceled + aborted
 
+    @staticmethod
+    def delete_old_observations(cutoff):
+        observations = Observation.objects.filter(start__lt=cutoff, end__lt=cutoff, state='CANCELED')
```
eheinrich: It is possible for an observation that is marked as canceled to have taken some data, in the case where the scheduler cancels a block, but a site lost connection and isn't able to update it's schedule, so it keeps on executing that observation that has been canceled. I think we should probably retain cancelled observations that have completed some time. What do you think?

jnation3406: Yeah I guess your right. I'll try excluding observations with configuration_statuses not in PENDING...

PR #55, opened 2019-07-20
Fix/optimization fixes

eheinrich:
The issue with the slow `/api/configurationstatus/` endpoint was its filter. For a filterset, if a field that is a foreign key is included, the default filter that is used is a [`ModelChoiceFilter`](https://django-filter.readthedocs.io/en/master/ref/filters.html#modelchoicefilter), which generates all the options for that field. So it's pretty slow on big tables.

The admin target change page was slow for a similar reason. The associated configuration is displayed on the page, and by default django will generate a dropdown with all options so that the user can update the associated configuration. Using [`raw_id_fields`](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.raw_id_fields) helps that.

### DIFF
--- a/observation_portal/observations/filters.py
+++ b/observation_portal/observations/filters.py
@@ -77,4 +77,4 @@ class ConfigurationStatusFilter(django_filters.FilterSet):
 
     class Meta:
         model = ConfigurationStatus
-        fields = '__all__'
+        fields = ('guide_camera_name',)
--- a/observation_portal/proposals/admin.py
+++ b/observation_portal/proposals/admin.py
@@ -91,10 +91,12 @@ class MembershipAdmin(admin.ModelAdmin):
 
 class ProposalInviteAdmin(admin.ModelAdmin):
     model = ProposalInvite
+    raw_id_fields = ['proposal']
 
 
 class ProposalNotificationAdmin(admin.ModelAdmin):
     model = ProposalNotification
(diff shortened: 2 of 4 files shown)


observation_portal/observations/filters.py
```diff
@@ -77,4 +77,4 @@ class ConfigurationStatusFilter(django_filters.FilterSet):
 
     class Meta:
         model = ConfigurationStatus
-        fields = '__all__'
+        fields = ('guide_camera_name',)
```
jnation3406: what about instrument name?

observation_portal/proposals/admin.py
```diff
@@ -91,10 +91,12 @@ def proposal_title(self, obj):
 
 class ProposalInviteAdmin(admin.ModelAdmin):
     model = ProposalInvite
+    raw_id_fields = ['proposal']
```
jnation3406: do raw_id_fields still give you a link to that admin object?

observation_portal/observations/filters.py
eheinrich: Instrument name is included as a filterable field since it is declared in the filterset a few lines up

observation_portal/proposals/admin.py
eheinrich: Yep!

PR #56, opened 2019-07-22
Fix target extra params

eheinrich:
Target extra params were not included in serialized responses. Also added vmag and radvel to request shim submission.


eheinrich:
Yes, neoexchange submits vmag via the shim, which was not getting through.

jnation3406:
oh yeah, I forgot NEOx coopted this NRES feature and started using it for tracking non sidereals... good catch

PR #58, opened 2019-07-28
allow eng to schedule on non-disabled instruments

jnation3406:


observation_portal/requestgroups/serializers.py
```diff
@@ -231,10 +231,14 @@ def validate_instrument_configs(self, value):
         return value
 
     def validate_instrument_type(self, value):
-        if value and value not in configdb.get_active_instrument_types({}):
+        is_staff = False
```
eheinrich: Could do `is_staff = self.context.get('request') or False`

jnation3406: just having a request context doesn't make it staff, the requests user must be staff. I think it might get ugly if I try to make that a single line...

PR #59, opened 2019-07-28
simplest changes for new instrument states

jnation3406:


jnation3406:
> I wonder if it would be better to add a parameter `include_states` alongside the `exclude_states` parameter, so you could do include_states=['SCHEDULABLE'] instead of having to pass through the longer exclude states list.

I'm not sure... we use the exclude_states because it is only excluding ['DISABLED'] in a lot of places. I don't know if it is worth the extra complexity of another parameter to shorten the 3 times where its all but one of the states...

PR #60, opened 2019-07-29
Add task for monthy time allocation reminders

Fingel:
Adds a task that, once a month, will send an email to every pi on a
current proposal with a summary of their allocated/used time. This is to
hopefully encourage users to use their time before it expires.

### DIFF
--- a/observation_portal/proposals/templates/proposals/timeallocationreminder.html
+++ b/observation_portal/proposals/templates/proposals/timeallocationreminder.html
@@ -1,0 +1,50 @@
+{% load i18n %}
+<!doctype html>
+<html lang="en">
+
+<head>
+    <title>Time Allocation Notification</title>
+</head>
+
+<body>
+<p>
+Dear {{ proposal.pi.first_name}} {{ proposal.pi.last_name }},
+</p>
+
+<p>
+The following table shows your network time allocation and usage of {{ proposal }} so far (through {% now "F j, Y" %}) for
+{{ proposal.current_semester }}.  Note that the semester ends on
+{{ proposal.current_semester.end|date:"F j, Y" }}, and unused time does not carry over.
+Also, be aware that the end of the semester generally sees higher contention for observing resources.
+</p>
+
+<table cellpadding=10>
+  <thead>
+    <tr>
+      <th></th><th colspan="3">Hours Allocated</th><th colspan="3">Hours Used</th>
+    </tr>
+    <tr>
+      <th></th><th>Queue</th><th>Time Critical</th><th>Rapid Response</th>
+      <th>Queue</th><th>Time Critical</th><th>Rapid Response</th>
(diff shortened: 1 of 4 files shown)


observation_portal/proposals/templates/proposals/timeallocationreminder.html
```diff
@@ -0,0 +1,50 @@
...
+</p>
+
+<table cellpadding=10>
+  <thead>
+    <tr>
+      <th></th><th colspan="3">Hours Allocated</th><th colspan="3">Hours Used</th>
+    </tr>
+    <tr>
+      <th></th><th>Queue</th><th>Time Critical</th><th>Rapid Response</th>
+      <th>Queue</th><th>Time Critical</th><th>Rapid Response</th>
```
jnation3406: Should these column headers have different names (like Queue Allocated, Queue Used)?

observation_portal/proposals/templates/proposals/timeallocationreminder.html
```diff
@@ -0,0 +1,50 @@
...
+  <tbody>
+    {% for ta in allocations %}
+    <tr>
+      <td>{{ ta.instrument_type }}</td>
+      <td>{{ ta.std_allocation }}</td>
+      <td>{{ ta.tc_allocation }}</td>
+      <td>{{ ta.rr_allocation }}</td>
+      <td>{{ ta.std_time_used|floatformat }}</td>
+      <td>{{ ta.tc_time_used|floatformat }}</td>
+      <td>{{ ta.rr_time_used|floatformat }}</td>
```
jnation3406: If the table is getting too wide, it could work to just return the time remaining for each observation type rather than the time allocated and time used

observation_portal/proposals/templates/proposals/timeallocationreminder.html
Fingel: There are 2 rows of headers, the top one says Hours Allocated, Hours Used. It looks good if you actually see it. There is a screenshot on the pivotal story.

jnation3406: Oh your right, I was having trouble visualizing it. It looks fine then

observation_portal/task_scheduler.py
```diff
@@ -19,5 +21,8 @@ def run():
         expire_access_tokens.send,
         CronTrigger.from_crontab('0 15 * * *')
     )
+    scheduler.add_job(
+        time_allocation_reminder.send,
+        CronTrigger.from_crontab('0 * 1 * *')  # monthly
```
eheinrich: I believe this schedule will send out an email once an hour on the first day of a month.

Fingel: Good catch. Fixed.

PR #67, opened 2019-08-15
Feature/update scicollab alloc

eheinrich:
As discussed, the allocations are by telescope name and the telescope options in the form are populated from configdb.

This is running on my machine for testing.


eheinrich:
Good call! I've updated all of the names in configdb. And yeah, I'll double check the values before deploy.

PR #73, opened 2019-08-21
Set the AWS_LOCATION for s3boto3 to be the mediafiles dir

jnation3406:
This is what was forgotten when we did away with the split_storage.py and static AWS storage. I think just setting this to the mediafiles dir will make the mediafiles be saved and loaded from that dir in the S3 bucket.


eheinrich:
I think this would also affect where the static files are placed... Maybe updating [this function](https://github.com/LCOGT/observation-portal/blob/master/observation_portal/sciapplications/models.py#L66) is a better way to go? That determines the [upload_to](https://docs.djangoproject.com/en/2.2/ref/models/fields/#filefield) parameter.

jnation3406:
static files aren't in s3 anymore? They are just in the normal staticfiles dir and nginx handles them I think

PR #74, opened 2019-08-26
Do not throw an exception on duration for old requests with unknown i types

Fingel:
I'd say this is a temporary solution at best. Better would be to cache the duration in the database so that the durations are persisted far into the future.

### DIFF
--- a/observation_portal/common/configdb.py
+++ b/observation_portal/common/configdb.py
@@ -15,6 +15,11 @@ class ConfigDBException(Exception):
     pass
 
 
+class InstrumentNotFoundException(Exception):
+    """Raised in an instrument is not found in the database """
+    pass
+
+
 class TelescopeKey(namedtuple('TelescopeKey', ['site', 'enclosure', 'telescope'])):
     """Generate the key of a telescope."""
     __slots__ = ()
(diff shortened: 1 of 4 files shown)


observation_portal/requestgroups/duration_utils.py
```diff
@@ -4,7 +4,7 @@
 import logging
 
 from observation_portal.proposals.models import TimeAllocationKey, Proposal, Semester
-from observation_portal.common.configdb import configdb
+from observation_portal.common.configdb import configdb, InstrumentNotFoundException
```
jnation3406: import not used?

observation_portal/common/configdb.py
```diff
@@ -553,7 +558,7 @@ def get_exposure_overhead(self, instrument_type, binning, readout_mode=''):
             # if the binning is not found, return the default binning (Added to support legacy 2x2 Sinistro obs)
             return default_mode['overhead'] + camera_type['fixed_overhead_per_exposure']
 
-        raise ConfigDBException(f'Instrument type {instrument_type} not found in configdb.')
+        raise InstrumentNotFoundException(f'Instrument type {instrument_type} not found in configdb.')
```
jnation3406: I'm thinking for this method and the next it might be better to return a default and log a warning that the instrument type wasn't found so a default was used. Then we shouldn't even need the later changes to catch this exception

jnation3406:
I'm closing this because it's probably pretty out of date now

PR #80, opened 2019-09-10
first round of deadlock fixes

jnation3406:
Here's a first round... There's still a few more update functions I want to tackle, and more testing on the dev cluster.

Basically I'm trying to reduce the sections that are in transaction.atomic, because I think that might be problematic. Also swapping out getting and then saving with a single select_for_update.update chain, which should be more effecient and accurate in some cases.

### DIFF
--- a/observation_portal/common/state_changes.py
+++ b/observation_portal/common/state_changes.py
@@ -1,5 +1,6 @@
 from django.utils import timezone
 from django.db import transaction
+from django.db.models import F
 from django.utils.translation import ugettext as _
 from django.core.cache import cache
 
@@ -12,12 +13,10 @@ from math import isclose, floor
 
 logger = logging.getLogger(__name__)
 
-
 REQUEST_STATE_MAP = {
-    'PENDING': ['COMPLETED', 'WINDOW_EXPIRED', 'CANCELED'],
-    'COMPLETED': [],
-    'WINDOW_EXPIRED': ['COMPLETED'],
-    'CANCELED': ['COMPLETED'],
+    'COMPLETED': ['PENDING', 'WINDOW_EXPIRED', 'CANCELED'],
+    'WINDOW_EXPIRED': ['PENDING'],
+    'CANCELED': ['PENDING']
 }
 
 TERMINAL_REQUEST_STATES = ['COMPLETED', 'CANCELED', 'WINDOW_EXPIRED']
@@ -41,7 +40,7 @@ class TimeAllocationError(Exception):
 
 
 def valid_request_state_change(old_state, new_state, obj):
-    if new_state not in REQUEST_STATE_MAP[old_state]:
+    if old_state not in REQUEST_STATE_MAP[new_state]:
         raise InvalidStateChange(_(f'Cannot transition from request state {old_state} to {new_state} for {obj}'))
 
 
@@ -65,7 +64,6 @@ def on_configuration_status_state_change(instance):
     update_request_group_state(instance.observation.request.request_group)
 
 
-@transaction.atomic
 def on_request_state_change(old_request, new_request):
     if old_request.state == new_request.state:
         return
@@ -92,15 +90,13 @@ def on_request_state_change(old_request, new_request):
                 modify_ipp_time_from_requests(ipp_value, [new_request], 'credit')
 
 
-@transaction.atomic
 def on_requestgroup_state_change(old_requestgroup, new_requestgroup):
     if old_requestgroup.state == new_requestgroup.state:
         return
     valid_request_state_change(old_requestgroup.state, new_requestgroup.state, old_requestgroup)
     if new_requestgroup.state in TERMINAL_REQUEST_STATES:
-        for r in new_requestgroup.requests.filter(state__in=['PENDING']):
-            r.state = new_requestgroup.state
-            r.save()
+        new_requestgroup.requests.filter(state__iexact='PENDING').update(
+            state=new_requestgroup.state, modified=timezone.now())
 
 
 def update_observation_state(observation):
@@ -161,8 +157,10 @@ def debit_ipp_time(request_group):
         total_duration_dict = request_group.total_duration
         for tak, duration in total_duration_dict.items():
             duration_hours = duration / 3600
-            time_allocations_dict[tak].ipp_time_available -= (ipp_value * duration_hours)
-            time_allocations_dict[tak].save()
+            ipp_difference = ipp_value * duration_hours
+            with transaction.atomic():
+                TimeAllocation.objects.select_for_update(of=('self')).filter(id=time_allocations_dict[tak].id).update(
+                    ipp_time_available=F('ipp_time_available') - ipp_difference)
     except Exception as e:
         logger.warning(_(
             f'Problem debiting ipp on creation for request_group {request_group.id} on proposal '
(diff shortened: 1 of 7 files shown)


observation_portal/common/state_changes.py
```diff
@@ -12,12 +13,10 @@
 
 logger = logging.getLogger(__name__)
 
-
 REQUEST_STATE_MAP = {
```
jnation3406: I inverted this mapping so it is new to old state now instead of old to new

observation_portal/common/state_changes.py
```diff
@@ -161,8 +157,10 @@ def debit_ipp_time(request_group):
         total_duration_dict = request_group.total_duration
         for tak, duration in total_duration_dict.items():
             duration_hours = duration / 3600
-            time_allocations_dict[tak].ipp_time_available -= (ipp_value * duration_hours)
-            time_allocations_dict[tak].save()
+            ipp_difference = ipp_value * duration_hours
+            with transaction.atomic():
+                TimeAllocation.objects.select_for_update(of=('self')).filter(id=time_allocations_dict[tak].id).update(
```
jnation3406: I'm not 100% sure the of=('self') does anything meaningful yet, but I am trying to avoid it locking related tables to the TimeAllocation (like proposal/semester). I saw evidence of those being locked before when looking through the dev database locks during deadlocks

observation_portal/requestgroups/serializers.py
```diff
@@ -656,45 +656,45 @@ class Meta:
...
-
-                for instrument_config_data in instrument_configs_data:
-                    rois_data = []
-                    if 'rois' in instrument_config_data:
-                        rois_data = instrument_config_data.pop('rois')
-                    instrument_config = InstrumentConfig.objects.create(configuration=configuration,
-                                                                        **instrument_config_data)
-                    for roi_data in rois_data:
-                        RegionOfInterest.objects.create(instrument_config=instrument_config, **roi_data)
+        with transaction.atomic():
```
jnation3406: for whatever reason, reducing the scope of this transaction.atomic so it didn't include the debit_ipp_time and the cache.set solved one of the deadlocking issues. Specifically, leaving the cache.set within the transaction.atomic didn't work, but I'm not sure why yet

eheinrich: Weird. I could see how excluding debit_ipp_time could help, cause that actually touches the db, but setting the cache doesn't... Does it sometimes take a long time to set the cache?

observation_portal/observations/viewsets.py
```diff
@@ -56,7 +56,7 @@ class ObservationViewSet(CreateListModelMixin, ListAsDictMixin, viewsets.ModelVi
     ordering = ('-id',)
 
     def get_queryset(self):
-        return observations_queryset()
+        return observations_queryset().prefetch_related('request__windows', 'request__location').distinct()
```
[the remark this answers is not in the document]
jnation3406: So I think the schedulerViewset doesn't need these and doesn't return them. It is just some validation for the observation serializer that uses the requests window and location on observation submission.

observation_portal/common/state_changes.py
```diff
@@ -92,15 +90,13 @@ def on_request_state_change(old_request, new_request):
...
-@transaction.atomic
 def on_requestgroup_state_change(old_requestgroup, new_requestgroup):
     if old_requestgroup.state == new_requestgroup.state:
         return
     valid_request_state_change(old_requestgroup.state, new_requestgroup.state, old_requestgroup)
     if new_requestgroup.state in TERMINAL_REQUEST_STATES:
-        for r in new_requestgroup.requests.filter(state__in=['PENDING']):
-            r.state = new_requestgroup.state
-            r.save()
+        new_requestgroup.requests.filter(state__iexact='PENDING').update(
```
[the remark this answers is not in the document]
jnation3406: Yeah your right, update doesn't call those handlers. I'll see if I can pipe the logic in a different way, but I think it's better to use update then use save and relay on the handlers. The handlers should be for user submitted updates only

observation_portal/requestgroups/serializers.py
jnation3406: no, not that I could see. It was literally the difference between deadlocking forever (or until the 5 minutes worker timeout happens) or not deadlocking. I'll spend some time looking into it a bit more when I've gotten the other things worked out

observation_portal/common/state_changes.py
```diff
@@ -92,36 +91,39 @@ def on_request_state_change(old_request, new_request):
...
-    valid_request_state_change(old_requestgroup.state, new_requestgroup.state, old_requestgroup)
+    valid_request_state_change(old_requestgroup_state, new_requestgroup.state, new_requestgroup)
     if new_requestgroup.state in TERMINAL_REQUEST_STATES:
-        for r in new_requestgroup.requests.filter(state__in=['PENDING']):
-            r.state = new_requestgroup.state
-            r.save()
+        for request in new_requestgroup.requests.filter(state__iexact='PENDING'):
+            request.state = new_requestgroup.state
+            request.save()
+        # new_requestgroup.requests.select_for_update(of=('self')).filter(state__iexact='PENDING').update(
```
eheinrich: Commented out code

observation_portal/common/state_changes.py
```diff
@@ -92,36 +91,39 @@ def on_request_state_change(old_request, new_request):
...
+def on_requestgroup_state_change(old_requestgroup_state, new_requestgroup):
+    if old_requestgroup_state == new_requestgroup.state:
         return
-    valid_request_state_change(old_requestgroup.state, new_requestgroup.state, old_requestgroup)
+    valid_request_state_change(old_requestgroup_state, new_requestgroup.state, new_requestgroup)
     if new_requestgroup.state in TERMINAL_REQUEST_STATES:
-        for r in new_requestgroup.requests.filter(state__in=['PENDING']):
-            r.state = new_requestgroup.state
-            r.save()
+        for request in new_requestgroup.requests.filter(state__iexact='PENDING'):
```
eheinrich: Is a case insensitive match required?

observation_portal/common/state_changes.py
eheinrich: Did you decide anything on this? Seems like it couldn't hurt but yeah, dunno exactly.

observation_portal/common/state_changes.py
jnation3406: no

observation_portal/common/state_changes.py
jnation3406: yeah I tested this out and it seems it is not necessary. After reading the docs a few more times, I think it only locks other rows and needs this if you do a 'select_related' on the queryset

observation_portal/common/state_changes.py
jnation3406: removed, don't think we need select_for_update here

observation_portal/common/state_changes.py
```diff
@@ -92,28 +91,28 @@ def on_request_state_change(old_request, new_request):
...
+def on_requestgroup_state_change(old_requestgroup_state, new_requestgroup):
+    if old_requestgroup_state == new_requestgroup.state:
         return
-    valid_request_state_change(old_requestgroup.state, new_requestgroup.state, old_requestgroup)
-    if new_requestgroup.state in TERMINAL_REQUEST_STATES:
-        for r in new_requestgroup.requests.filter(state__in=['PENDING']):
-            r.state = new_requestgroup.state
-            r.save()
+    valid_request_state_change(old_requestgroup_state, new_requestgroup.state, new_requestgroup)
+    if new_requestgroup.state in ['CANCELED', 'WINDOW_EXPIRED']:
```
jnation3406: changed this because I think it should only update the Request state if the RG was not COMPLETED. We don't want PENDING requests set to COMPLETED...

eheinrich: Yeah, good call

observation_portal/common/state_changes.py
```diff
@@ -265,43 +280,49 @@ def update_request_states_for_window_expiration():
...
-                        req = Request.objects.select_for_update().get(pk=request.id)
-                        if req.state == 'PENDING':
-                            req.state = 'WINDOW_EXPIRED'
-                            states_changed = True
+                        if Request.objects.select_for_update().filter(pk=request.id, state='PENDING').update(state='WINDOW_EXPIRED'):
+                            any_states_changed = True
                             request_states_changed = True
-                            req.save()
+                            request_state_changed = True
+                    if request_state_changed:
```
eheinrich: I think `request_state_changed ` might not have anything assigned to it at this point.

jnation3406: you're right! I didn't test DIRECT code path specifically

PR #95, opened 2019-11-04
Feature/enable multiple instrument configs

eheinrich:
Updated the serializers and the compose page. Added validation checks for what we don't want to allow further down in the stack. 

Restricting SCRIPT and SKY_FLAT type configurations from having multiple instrument configurations won't matter too much now since those are generally directly submitted, but those checks will come in handy for when we re-enable validating configurations for directly submitted observations.

### DIFF
--- a/observation_portal/requestgroups/serializers.py
+++ b/observation_portal/requestgroups/serializers.py
@@ -250,13 +250,6 @@ class ConfigurationSerializer(serializers.ModelSerializer):
         default_modes = configdb.get_default_modes_by_type(instrument_type)
         guiding_config = data['guiding_config']
 
-        # Validate the number of instrument configs if this is not a SOAR instrument
-        # TODO: remove this check once we support multiple instrument configs
-        if len(data['instrument_configs']) != 1 and 'SOAR' not in instrument_type.upper():
-            raise serializers.ValidationError(_(
-                'Currently only a single instrument_config is supported. This restriction will be lifted in the future.'
-            ))
-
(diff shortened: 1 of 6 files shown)


observation_portal/requestgroups/serializers.py
```diff
@@ -228,6 +228,8 @@ class Meta:
     def validate_instrument_configs(self, value):
         if [instrument_config.get('fill_window', False) for instrument_config in value].count(True) > 1:
             raise serializers.ValidationError(_('Only one instrument_config can have `fill_window` set'))
+        if len(set([instrument_config.get('rotator_mode', '') for instrument_config in value])) > 1:
+            raise serializers.ValidationError(_('Rotator modes within the same configuration must be the same'))
```
jnation3406: is this a limitation of the site software implementation?

eheinrich: Yeah, site software will use rotator mode in the first inst config and ignore the rest of them. So, technically, we could allow a user to pass in different rotator modes, but I think that would just end up being confusing if they don't get what they thought they asked for...

eheinrich: From Mark: "We do not support multiple rotator modes or angles. The mode and angle in the first Instrument Configuration will be used for all associated Instrument Configurations. This is because we do not want to introduce an axis slew between configurations which would blow-up our overheads estimate".

jnation3406: interesting...

PR #96, opened 2019-12-02
add Time Critical option in compose page

jnation3406:


static/js/components/requestgroup.vue
```diff
@@ -78,11 +78,13 @@
               label="Mode"
               field="observation_type" v-on:input="update"
               desc="Rapid Response (RR) requests bypass normal scheduling and are executed immediately. 
-                    This mode is only available if a proposal was granted RR time."
+                    Time Critical (TC) requests are given a large fixed priority that will beat any Normal requests.
+                    These mode are only available if a proposal was granted RR or TC time."
```
eheinrich: "... that will beat any requests that use default queue scheduling." or something like that

"These modes ..."

PR #97, opened 2019-12-04
Config duration

jnation3406:


static/js/components/instrumentconfig.vue
```diff
@@ -39,20 +39,8 @@
               field="exposure_count" 
               type="number"
               :errors="errors.exposure_count" 
-              desc="Clicking the 'Fill' button will set the exposure count to the number of 
-                    exposures (including overheads) that will fit in the largest observing window. 
-                    This button is disabled until the entire observation request has passed validation."
```
eheinrich: Might be a good idea to have this note in where the new description for the fill window button is.

PR #100, opened 2020-01-24
Fix/ repeat duration

eheinrich:
Use the repeat_duration to figure out percent completed for repeat type configurations

### DIFF
--- a/observation_portal/requestgroups/request_utils.py
+++ b/observation_portal/requestgroups/request_utils.py
@@ -122,26 +122,32 @@ def get_airmasses_for_request_at_sites(request_dict, is_staff=False):
 
 
 def exposure_completion_percentage(configuration_statuses):
-    total_exposure_time = 0
-    completed_exposure_time = 0
+    total_time = 0
+    completed_time = 0
     for configuration_status in configuration_statuses:
-        if hasattr(configuration_status, 'summary'):
-            completed_exposure_time += configuration_status.summary.time_completed
-        configuration_exposure_time = 0
-        if (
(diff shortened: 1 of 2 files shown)


observation_portal/requestgroups/request_utils.py
```diff
@@ -122,20 +122,32 @@ def get_airmasses_for_request_at_sites(request_dict, is_staff=False):
...
+                completed_time += (
+                        configuration_status.summary.end - configuration_status.summary.start
+                ).total_seconds()
+            configuration_total_time = configuration_status.configuration.repeat_duration
+        else:
+            if has_summary:
+                completed_time += configuration_status.summary.time_completed
+            for instrument_config in configuration_status.configuration.instrument_configs.all():
+                configuration_total_time += instrument_config.exposure_count * instrument_config.exposure_time
+        total_time += configuration_total_time
```
jnation3406: I might be misreading this, but it seems like you could just have one variable here that is zero in the beginning and adds onto the total time the whole way through? I probably wrote it the other way for no good reason before

eheinrich: Do you mean doing something like adding on the `total_time` variable directly, and getting rid of the intermediate `configuration_total_time` variable?

jnation3406: yeah that's what I meant. It seemed like an unnecessary extra variable

eheinrich: :+1: Agreed, done!

PR #101, opened 2020-01-25
Update percent complete for repeat types in tooltip

eheinrich:


static/js/request_detail.vue
```diff
@@ -410,27 +410,45 @@
...
             for (let csIndex in that.observationData[observationIdx].configuration_statuses) {
               let configurationStatus = that.observationData[observationIdx].configuration_statuses[csIndex];
-              if (!_.isEmpty(configurationStatus.summary)) {
-                time_completed += configurationStatus.summary.time_completed;
-                if (fail_reason === '' && configurationStatus.summary.reason !== 'N/A') {
-                  fail_reason = configurationStatus.summary.reason;
-                }
-              }
+              // Find the configuration that goes with this configuration status
+              let configuration = that.request.configurations[0];
```
jnation3406: should we maybe leave this undefined? If we don't find a configuration corresponding to the configuration status, I think I would rather raise an error in the console then have it appear fine but with the wrong configuration...

eheinrich: Yes, I will do that.

PR #105, opened 2020-02-21
Clean up blocks and userrequests

eheinrich:
In preparation for turning off the request and lake shims, anything having to do with userrequests and blocks should have been removed.


observation_portal/requestgroups/test/test_api.py
```diff
@@ -2145,100 +2142,6 @@ def test_cannot_cancel_completed_rg(self, modify_mock):
         self.assertEqual(Request.objects.get(pk=completed_r.id).state, 'COMPLETED')
 
 
-@patch('observation_portal.common.state_changes.modify_ipp_time_from_requests')
```
eheinrich: I don't remember why we kept these commented-out tests... They are testing the `isDirty` endpoint, which we replaced with the `last_changed` endpoint, which is tested elsewhere. And they also check some request state updates, but state transitions are also tested elsewhere. Let me know if you see anything that this would have been testing that I should add a test for!

jnation3406: pretty sure we commented them early on when the pond block stuff was removed, just to remember to cover those cases later. I don't see anything special in them now, so I think it's fine to delete

PR #106, opened 2020-03-02
fix hour_angle equal 0 validation bug

phycodurus:
with the deprication of dict.has_key() from Python3, the nice
way to check key inclusion is 'key' in dict.


phycodurus:
This PR is not quite ready as several tests are failing. A workaround has been implemented in calibration_utils for the time being.

PR #112, opened 2020-05-08
Feature/public docker images

eheinrich:
This PR moves docker image builds from being run by LCO's internal infrastructure to being run by Travis and pushed to Docker Hub.

The behavior maintains what our current default Jenkins pipeline does. When a new git tag is created, the image will be tagged with that git tag. If it is a normal commit, the image will be tagged with the git commit. As far as I can tell there is no limit to how many tags we can house in Docker Hub for open projects. Let me know if you think this is excessive.

The docker image will only be built for branches in this repository, not from forks. Even if a fork removed the restriction that docker builds only happen on this repo, they would not have access to our docker hub credentials so it would not work to push to our repo.

The images are pushed to the `observatorycontrolsystem` organization on Docker Hub: https://hub.docker.com/u/observatorycontrolsystem


eheinrich:
Another thing I should mention is that I looked into using GitHub Packages to host our docker images, and while that does work, there is a requirement that when you pull an image from GHP, you have to be authenticated, even if the image is public. I felt this was the wrong way to go for an open source project because I wanted it to be as easy as possible to use our images, so I went with Docker Hub. It looks like other [people also feel the same way](https://github.community/t5/GitHub-Actions/docker-pull-from-public-GitHub-Package-Registry-fail-with-quot/td-p/32782)!

PR #120, opened 2020-05-13
Muscat ui elements

jnation3406:
I leveraged the extra_params requirements we already had in place for GenericModes from configdb to do most of the validation for us. The only specific validation is with checking that the exposure_time_* fields are non-negative numbers (since it doesn't do any type checking by default). I think we could integrate in some basic type checking automatically with the required params in configdb as well, but not sure if its worth doing now, or wait for a bigger overhaul of configdb/obs portal to include it, since it should help a lot with generalizing stuff

### DIFF
--- a/observation_portal/requestgroups/serializers.py
+++ b/observation_portal/requestgroups/serializers.py
@@ -36,8 +36,14 @@ class ModeValidationHelper:
         self._instrument_type = instrument_type
         self._default_modes = default_modes
         self._modes = modes
-        self._mode_key = 'rotator_mode' if self._mode_type == 'rotator' else 'mode'
+        self._set_mode_key(self._mode_type)
         self._modes_by_code = {}
+
+    def _set_mode_key(self, mode_type):
+        if mode_type == 'rotator' or mode_type == 'exposure':
+            self._mode_key = mode_type + '_mode'
+        else:
+            self._mode_key = 'mode'
 
     def _possible_modes(self) -> list:
         possible_modes = []
@@ -48,18 +54,18 @@ class ModeValidationHelper:
             possible_modes.extend(self._modes[self._mode_type]['modes'])
         return possible_modes
 
-    def _unavailable_msg(self, config: dict) -> str:
+    def _unavailable_msg(self, mode_value: str, config: dict) -> str:
         if self._mode_type in self._modes:
-            if not config[self._mode_key].lower() in [m['code'].lower() for m in self._modes[self._mode_type]['modes']]:
+            if not mode_value.lower() in [m['code'].lower() for m in self._modes[self._mode_type]['modes']]:
                 return (
-                    f'{self._mode_type.capitalize()} mode {config[self._mode_key]} is not available for '
+                    f'{self._mode_type.capitalize()} mode {mode_value} is not available for '
                     f'instrument type {self._instrument_type}'
                 )
         return ''
 
-    def _missing_fields_msg(self, config) -> str:
+    def _missing_fields_msg(self, mode_value: str, config: dict) -> str:
         missing_fields = []
-        mode = configdb.get_mode_with_code(self._instrument_type, config[self._mode_key], self._mode_type)
+        mode = configdb.get_mode_with_code(self._instrument_type, mode_value, self._mode_type)
         if 'required_fields' in mode.get('params', {}):
             for field in mode['params']['required_fields']:
                 if 'extra_params' not in config or field not in config['extra_params']:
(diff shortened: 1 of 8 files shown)


observation_portal/requestgroups/serializers.py
```diff
@@ -160,6 +166,26 @@ def validate(self, data):
             raise serializers.ValidationError(_(
                 'Currently only square binnings are supported. Please submit with bin_x == bin_y'
             ))
+
+        extra_param_max_exp_time = 0
+        for field, value in data.get('extra_params', {}).items():
```
jnation3406, line 171: This is for the requirement that exposure_time == the max of the individual exposure times

observation_portal/requestgroups/serializers.py
```diff
@@ -160,6 +166,26 @@ def validate(self, data):
...
+        if extra_param_max_exp_time > 0:
+            data['exposure_time'] = extra_param_max_exp_time
+
+        return data
+
+    def to_internal_value(self, data):
+        data = super().to_internal_value(data)
+        for field, value in data.get('extra_params', {}).items():
+            try:
+                data['extra_params'][field] = float(value)
```
jnation3406, line 186: This is so that extra params that are numbers are stored as such in the database json field (and then are retrieved that way as well)

observation_portal/requestgroups/serializers.py
```diff
@@ -421,6 +450,35 @@ def validate(self, data):
                             ))
                         )
 
+            # Validate the exposure modes
+            if 'exposure' in modes:
```
jnation3406, line 454: This is a new GenericMode type in configdb, and the two options for muscat both have the exposure_time_* values as required params, so this makes sure they exist. It also checks the value of the exposure_mode is valid, and if not set sets the default value.

observation_portal/requestgroups/serializers.py
```diff
@@ -421,6 +450,35 @@ def validate(self, data):
...
+                        instrument_config['extra_params']['exposure_mode'] = exposure_mode_to_set['mode']['code']
+
+                exposure_mode_error_msg = exposure_mode_validation_helper.get_mode_error_msg(
+                    instrument_config['extra_params']['exposure_mode'],
+                    instrument_config
+                )
+                if exposure_mode_error_msg:
+                    raise serializers.ValidationError(_(exposure_mode_error_msg))
+
+            # Validate the exposure times for MUSCAT are positive numbers
```
jnation3406, line 471: This extra stuff is to validate the exposure_time_* are non-negative numbers. We should maybe make this type of validation automatic in the ModeValidationHelper by defined a schema to define type of required fields in configdb, like `{"required_fields": ["exposure_time_g[float]", "exposure_mode[string]"]}` or something like that

observation_portal/requestgroups/serializers.py
eheinrich: Sounds reasonable to add to other serializers.

observation_portal/requestgroups/serializers.py
```diff
@@ -421,6 +450,35 @@ def validate(self, data):
...
                         )
 
+            # Validate the exposure modes
+            if 'exposure' in modes:
+                exposure_mode_validation_helper = ModeValidationHelper('exposure', instrument_type, default_modes,
+                                                                       modes)
+                if exposure_mode_validation_helper.mode_is_not_set(instrument_config['extra_params']):
+                    exposure_mode_to_set = exposure_mode_validation_helper.get_mode_to_set()
+                    if exposure_mode_to_set['error']:
+                        raise serializers.ValidationError(_(rotator_mode_to_set['error']))
```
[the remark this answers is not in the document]
jnation3406: yup

jnation3406:
Yeah I see when exposure_time_* is not set, there is an error that isn't displayed. Not sure exactly where to put that since its just a general Configuration error, but maybe I can add it to the exposure_mode since it is it's required fields that are missing

observation_portal/requestgroups/serializers.py
```diff
@@ -421,6 +450,35 @@ def validate(self, data):
...
+                exposure_mode_error_msg = exposure_mode_validation_helper.get_mode_error_msg(
+                    instrument_config['extra_params']['exposure_mode'],
+                    instrument_config
+                )
+                if exposure_mode_error_msg:
+                    raise serializers.ValidationError(_(exposure_mode_error_msg))
+
+            # Validate the exposure times for MUSCAT are positive numbers
+            if instrument_type.upper() == '2M0-SCICAM-MUSCAT':
+                exposure_fields = ['exposure_time_g', 'exposure_time_r', 'exposure_time_i', 'exposure_time_z']
```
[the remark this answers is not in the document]
jnation3406: if we update configdb to handle types automatically, then this code will go away, so I guess I'll just move forward with that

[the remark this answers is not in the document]
eheinrich: Sounds good 👍

eheinrich:
The error that I am seeing is actually that the exposure_time on an instrument_config cannot be null. I saved a draft request in the development portal named "muscat validation error" that has that error.

observation_portal/requestgroups/serializers.py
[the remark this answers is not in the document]
jnation3406: so I have another branch for this, but i think we should release this first and then spend a bit more time on the other branch later

observation_portal/requestgroups/serializers.py
```diff
@@ -421,6 +452,44 @@ def validate(self, data):
...
+                for exposure_field in exposure_fields:
+                    try:
+                        exposure_time = float(instrument_config['extra_params'][exposure_field])
+                        if exposure_time < 0:
+                            raise serializers.ValidationError(_(f'Muscat instrument extra_param {exposure_field} must be a positive number.'))
+                    except ValueError:
+                        raise serializers.ValidationError(_(f'Muscat instrument extra_param {exposure_field} must be a positive number.'))
+            # This applies the exposure_time requirement only to non-muscat instruments.
+            # I wish I could figure out a better way to do this, but it needs info about the instrument_type which is a level up.
```
eheinrich, line 484: I don't quite understand - I think `instrument_type` is actually available from here. What other instrument_type info does this section need?

jnation3406: sorry, I meant that to validate the exposure time extra params, we need to know that the instrument type is MUSCAT, which is only available at the Configuration level. So even though it would be better to validate the Instrument Config extra params at that level/serializer, we have to do it in the Configuration one because that's where we have the instrument type

eheinrich: Ah, gotcha. I also wish there was a better way to do this, but nothing is coming to me right away.

PR #130, opened 2020-06-10
Refactor the ModeValidatorHelper a bit to try to simplify usage

jnation3406:
This passes all tests so I think it does what was done before and is a little simpler, but I'm not opposed to taking another pass or two at it because I still think it could be simplified more. 

It basically contains most of the mode validation state inside the ModeValidationHelper class, and the entry point to use it is just to call the .validate() method, and then depending on the result, get the classes .validated_config_dict, or get its .validation_errors string to raise an exception. What are your thoughts?

### DIFF
--- a/observation_portal/requestgroups/serializers.py
+++ b/observation_portal/requestgroups/serializers.py
@@ -48,22 +48,20 @@ def cerberus_validation_error_to_str(validation_errors: dict) -> str:
 
 class ModeValidationHelper:
     """Class used to validate GenericModes of different types defined in ConfigDB"""
-    def __init__(self, mode_type, instrument_type, modes_group, config_dict, mode_key='mode', is_extra_param_mode=False):
+    def __init__(self, mode_type, instrument_type, modes_group, mode_key='mode', is_extra_param_mode=False):
         self._mode_type = mode_type.lower()
         self._instrument_type = instrument_type
-        self._config_dict = config_dict
         self._modes_group = modes_group
         self._mode_key = mode_key
         self.is_extra_param_mode = is_extra_param_mode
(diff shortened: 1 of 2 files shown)


observation_portal/requestgroups/serializers.py
```diff
@@ -302,22 +304,16 @@ def validate(self, data):
...
-            guiding_config['mode'], guiding_config
-        )
-        data['guiding_config'] = guiding_config
-        if guide_mode_error_msg:
-            raise serializers.ValidationError(_(guide_mode_error_msg))
+        guide_validation_helper = ModeValidationHelper('guiding', instrument_type, modes['guiding'])
+
+        guiding_config = guide_validation_helper.validate(guiding_config)
+        if not guiding_config.get('mode'):
+            # Guiding modes have an implicit default of OFF (we could just put this in all relevent validation_schema)
```
eheinrich, line 311: Already as it is, I think the mode will be set at this point already because the modes in configdb generally will all have defaults set. Is this what you are referring to? I do agree that getting the default from configdb sounds better than setting this here in the code.

jnation3406: At least for our unit tests configdb data that wasn't the case, so I left this in since this is what the code used to do. I'm not sure if we should remove it, because on the one hand it is useful to have a sane default (guiding mode now defaults to empty string), but on the other hand it would be nice to remove this and rely entirely on configdb being setup 'correctly'

eheinrich: Do you remember why we set the default guide mode to an empty string? I'm wondering if we can set it to `GuidingConfig.OFF` instead.

PR #140, opened 2020-07-07
Add muscat time to semester detail page

jnation3406:


jnation3406:
it looks like we still have some wonkiness in the email tests that happens depending on the order they are run in.

PR #147, opened 2020-08-16
Feature/add accounts pages

eheinrich:
Changes:
- Remove the `/accounts/profile/` page which has previously been used to display and update a user's profile info. Instead the `/api/profile/` endpoint has been expanded such that it can be used to update profile information.

- Remove the `/accounts/acceptterms/` page. Instead the terms can be accepted by using the `/api/profile/` endpoint with the data `{"profile": {"accept_terms": true}}`. I figured it made sense to have the accept terms functionality happen here as the `terms_accepted` field is a profile field, but let me know if you disagree. The middleware to redirect a user that has not yet accepted the terms to the acceptterms page has been removed. This is now done in the frontend.

- Remove the `/accounts/revoketoken/` and `/accounts/removalrequest/` pages and replace them with API endpoints to do the same thing.

- Add `/accounts/loggedinstate/` page and set `LOGIN_REDIRECT_URL` to it. The default value for `LOGIN_REDIRECT_URL` is actually `/accounts/profile/`, which strangely is not a page that is provided by django. I've removed the setting for `LOGOUT_REDIRECT_URL` as that sensibly defaults to a page that lets the user know they have successfully logged out.

- Add `/accounts/` to `CORS_URLS_REGEX` in the settings file so that the frontend can get those endpoints with AJAX calls.

- Removed LCO specific things from accounts forms.


observation_portal/accounts/serializers.py
```diff
@@ -1,28 +1,49 @@
...
 
+    def validate_staff_view(self, staff_view):
+        user = self.context.get('request').user
+        is_staff = user and user.is_authenticated and user.is_staff
+        if staff_view and not is_staff:
+            raise serializers.ValidationError(_('Must be staff to set staff_view'))
+        return staff_view
+
+    def validate_accept_terms(self, accept_terms):
+        if not accept_terms:
```
jnation3406, line 29: I guess this feels a bit strange that they have to submit `"accept_terms": True` when you can't submit it as False. Maybe it would be better to just have a /api/profile/accept_terms endpoint that sets the field to True, and then the field has no other mechanism of being updated? Maybe that is too convoluted though and this is fine as is....

eheinrich: I made the change... As far as I could tell, `APIView`s don't allow me to add another endpoint on with the `action` decorator like we do in our ViewSets, so I added another `APIView` solely for accepting the terms and explicitly mapped it to `/api/profile/accept_terms`.

eheinrich:
That's right. The plan is to merge changes into the `feature/separate-frontend` branch, which won't get merged in until the switch.

PR #149, opened 2020-08-24
Add a filter to get a requestgroup by a request id

eheinrich:
The separated frontend needs to be able to get the requestgroup given a request ID for the request detail page.

### DIFF
--- a/observation_portal/requestgroups/filters.py
+++ b/observation_portal/requestgroups/filters.py
@@ -31,7 +31,7 @@ class RequestGroupFilter(django_filters.FilterSet):
             'modified': 'Last Update'
         }
     )
-    request_id = django_filters.CharFilter(field_name='requests__id')
+    request_id = django_filters.NumberFilter(field_name='requests__id')
 
     class Meta:
         model = RequestGroup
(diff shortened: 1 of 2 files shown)


observation_portal/requestgroups/filters.py
```diff
@@ -31,12 +31,13 @@ class RequestGroupFilter(django_filters.FilterSet):
             'modified': 'Last Update'
         }
     )
+    request_id = django_filters.CharFilter(field_name='requests__id')
```
jnation3406, line 34: should this be an integer filter instead of char filter - I think the ids are autofields, which are integer fields...

jnation3406 (approved):

observation_portal/requestgroups/filters.py
eheinrich: Hmm yeah maybe... we do the same thing in the [observation filters](https://github.com/observatorycontrolsystem/observation-portal/blob/master/observation_portal/observations/filters.py#L48). Do you happen to remember why we went with a charfield there?

jnation3406: nope, I see I did that, but I probably just didn't know and it worked? It should probably be an integer filter too.

eheinrich: Ok I'll change them

PR #153, opened 2020-09-02
change the pdf upload page limit requirement to a recommendation

jnation3406:


eheinrich (commented):
Another thing that we wanted to add to the text is the size limit of the uploaded file

eheinrich (approved):
Looks good to me!

PR #157, opened 2020-09-25
Style fixes

zemogle:
As discussed with @eheinrich

### DIFF
--- a/observation_portal/accounts/templates/auth/user_form.html
+++ b/observation_portal/accounts/templates/auth/user_form.html
@@ -4,7 +4,7 @@
 {% block content %}
 <div class="row">
     <div class="col-md-8">
-        <h2>Settings</h2>
+        <h3>Settings</h3>
         <form class="form" method="post" action="">
         {% csrf_token %}
         {% bootstrap_form user_form %}
@@ -15,7 +15,7 @@
         </form>
     </div>
     <div class="col-md-4">
-        <h2>API Key</h2>
+        <h3>API Key</h3>
         <p>
         The following key may be used to authenticate when using the
         <a href="{% url 'api:api-root' %}">API</a>.
@@ -31,13 +31,13 @@
             {% csrf_token %}
             <input type="submit" class="btn btn-danger" value="Revoke Token" />
         </form>
-        <h2>API Quota</h2>
+        <h3>API Quota</h3>
         <p>The observations API is rate limited so that it remains available to all users in the event
         a single user makes too many requests in too rapid a succession, causing denial of service to
         other clients.</p>
         <p>The rate limit is applied on a 24 hour rolling window. Your current quota is:</p>
         <p><strong>{{ user.profile.api_quota.used }} out of {{ user.profile.api_quota.allowed }}</strong></p>
-        <h2>Account Removal</h2>
+        <h3>Account Removal</h3>
         <p>You may request that your account be deleted.<p>
         <a href="{% url 'account-removal' %}" title="Request account removal">Request account removal</a>
     </div>
--- a/observation_portal/proposals/templates/proposals/proposal_detail.html
+++ b/observation_portal/proposals/templates/proposals/proposal_detail.html
@@ -21,7 +21,7 @@
               <sup><a
                 data-toggle="tooltip"
                 data-placement="bottom"
-                title="You will recieve notifications whenever a
+                title="You will receive notifications whenever a
                 requested observation on this proposal is completed.
                 ">
                 ?</a></sup>
--- a/observation_portal/proposals/templates/proposals/proposal_detail.html
+++ b/observation_portal/proposals/templates/proposals/proposal_detail.html
@@ -18,7 +18,7 @@
       </div>
       <div class="col-md-4">
             <strong>Email Notifications
-              <sup><a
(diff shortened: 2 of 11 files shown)


eheinrich:
The changes for this PR that we had talked about were:
- ✅ **Buttons should change color on hover**
- ✅ **The text of a navigation item in the header should be high contrast when hovering over it** 
- ✅ **The header should be the same on each page**. There is still some other CSS that is making the header appear smaller on certain pages, but I will address this in the separated frontend.
- ✅ **Restructure the header to make better use of the space**
- ✅ **Pagination buttons styling, as well as a few other links, were using the wrong shade of blue**
- ✅ **Use consistent button styling instead of the mixture of purely bootstrap-styled buttons and new custom styled buttons**. When we talked about this, we were looking at buttons on the profile page, which look good now! We also looked at buttons on the request detail page, as well as the filter list dropdown button on the home page, which are all still using the default bootstrap styling. I am happy to address this in the separated frontend.
- **The hover-over highlighter yellow on the quick navigation items on the home page only covered part of the icons in the links.** The solution in this PR of using a button group does indeed solve that issue! However, I'm not sure that it fits with the style/flow/feel of the rest of the page. The side navigation is split up into sections, with each section including a title and content, and where all links are styled as links. Removing the title from the quicknav section and using a button group disrupts that consistency and as such I personally find it less visually pleasing. I think even just going back to the original quick nav styling, and removing the `fa-2x` class from the icons used in the links actually works out better. What do you think?

One other comment from me - what do you think about making the question mark next to "Telescope availability history" on the home page blue? On my screen the question mark shows up on the next line, so it ends up looking like the title says "Telescope availability history?". Changing the color would make it more obviously stand out as a helper tooltip and as separate from the title.

I've deployed this branch to our development environment and shared it for feedback on the changes to the proposals detail page and the apply page. The only change request was for the apply page:

> I'm fine with removing the description of the three types of proposals, because that information is available on the Time Allocation Process page, which we link to. I'm not fine with removing the text that begins "If you are composing a proposal for the LCO network for the first time," and includes links to other information-rich pages. Would you please restore that paragraph?

zemogle:
> The hover-over highlighter yellow on the quick navigation items on the home page only covered part of the icons in the links. The solution in this PR of using a button group does indeed solve that issue! However, I'm not sure that it fits with the style/flow/feel of the rest of the page. The side navigation is split up into sections, with each section including a title and content, and where all links are styled as links. Removing the title from the quicknav section and using a button group disrupts that consistency and as such I personally find it less visually pleasing. I think even just going back to the original quick nav styling, and removing the fa-2x class from the icons used in the links actually works out better. What do you think?

 - I actually disagree. With my developer hat on, 'submit observation', 'manage proposals' and 'help' are all calls to action. Calls to action should be buttons.
- With my user hat on: The heading "Quick Navigation" is clutter and doesn't provide any value. Likewise, why  "Need More Information?" then a link in a paragraph text. A button saying "Help" not only has an obvious purpose but is much cleaner.
- I don't believe it interrupts the flow because everything else in that side menu is informational, not a call to action, so headings are appropriate.

If I haven't convinced you maybe the best option is to canvas some of the user community?

**Submit Proposal Page**
I have added back the requested information for first time proposers. I would put out that this content is out of date and contains 4 links to "lcogt.net". Since textual changes like this have a significant overhead, I strongly recommend that this paragraph is trimmed down in a future release and points to the main LCO website, which is easier and faster to edit.

**?**
Good idea. I've made the other tooltips I'd added blue as well.

eheinrich:
> I actually disagree. With my developer hat on, 'submit observation', 'manage proposals' and 'help' are all calls to action. Calls to action should be buttons.

I not sure that I agree with this actually... 'Manage proposals' and 'submit observation' are actions, but the purpose of those links is to navigate you to the pages where you can then perform those actions. I think there are definitely places where it makes sense to style a link as a button, however in this case, leaving them styled as links keeps their purpose clear. And again, leaving them as styled as links keeps them more consistent with the style of the rest of the page. Also, we have 'manage proposals' and 'submit observation' navigation links in the header styled as links as well, which I would not want to style as buttons.

> With my user hat on: The heading "Quick Navigation" is clutter and doesn't provide any value. Likewise, why "Need More Information?" then a link in a paragraph text. A button saying "Help" not only has an obvious purpose but is much cleaner.

I have no objection to moving the 'Help' link into the quick navigation section and getting rid of the "Need More Information" section. I personally don't think that the "Quick Navigation" title is clutter, but I can see that a case can be made to remove it. I **do** think there is value in having titles on separate sections of content though, both for clarity and in this case for keeping consistent with the style of the rest of the page.

> If I haven't convinced you maybe the best option is to canvas some of the user community?

Indeed I am not convinced... Seeing as this PR should be merged in relatively soon, before proposal submission heats up, I think it might be best for us to park this discussion for now, revert the quicknav back to the previous style, and then have more of a discussion when time permits. Would that be ok with you?

zemogle:
I definitely don't agree with your point about needing the heading "Quick Navigation" or the point about calls to action in navbar links. However I have fixed the issue with the highlight on those links and reverted the heading, so you can deploy this branch.

zemogle:
and I put the contraversial buttons back to links too 😉

PR #160, opened 2020-11-02
Allow additional muscat guiding modes

eheinrich:
We have only ever had guide modes `ON` and `OFF`, which mapped to user selectable guiding options `On`, `Off`, and `Optional`. We have recently added additional guide modes for muscat. We had previously just set the `Optional` guide option for muscat requests on the compose page and hidden other options, since that was the only mode that made sense. Since adding other modes, we need to show these new modes in the guiding dropdown on the compose page so that a user can select them.

The changes in this PR:
- Allow selection of additional muscat guide modes.
- I tried to clean up the code a little bit. User selectable guide mode options are different from the guiding_config fields that are set, but they do map in a predictable manner. I added functions to map from guiding option <==> guiding_config.

### DIFF
--- a/static/js/components/configuration.vue
+++ b/static/js/components/configuration.vue
@@ -517,7 +517,7 @@
       selectedImagerGuidingOption: function(value) {
         this.setGuidingFields(value);
       },
-      guideModeOptions: function(value) {
+      guideModeOptions: function() {
         // The selected guide mode for spectrographs is already updated elsewhere
         if (this.datatype !== 'SPECTRA') {
           let newGuideOption = this.getNewGuideOption(this.selectedImagerGuidingOption);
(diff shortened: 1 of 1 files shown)


jnation3406 (approved):
This seems good for now. I think long term we may want to split off the optional into a separate field in the UI That is hidden if you select mode "OFF". And if we ever have non-OFF modes that can't be optional (I guess spectrographs are like this already?) then we should probably add just add the allowable values of "optional" to the Cerberus validation schema for the guiding modes so it can automatically validate...

eheinrich:
That's a great idea. Having a separate field in the UI would certainly cut down on the frontend code complexity. I'll capture those ideas in a story so we don't forget.

PR #163, opened 2020-12-01
Feature/sciapp pages

eheinrich:
- Add a viewset for science applications API endpoints, replacing the forms that had been used.
- Combined PDF generation has been moved entirely to the new frontend and removed from here.
- Remove all the old views that provided science application pages. Since these are the final set of pages that are being replaced, also removed other remaining frontend code and configuration.


observation_portal/common/configdb.py
```diff
@@ -312,21 +312,21 @@ def get_instrument_names(
                 instrument_names.add(instrument['code'].lower())
         return instrument_names
 
-    def get_instrument_types_per_telescope_name(self, exclude_states=None) -> dict:
-        """Get a set of instrument types.
+    def get_telescope_name_by_instrument_types(self, exclude_states=None) -> dict:
+        """Get a mapping of instrument type to telescope name.
 
-        Instrument types are returned by telescope name (1 meter, 2 meter, etc...)
+        Telescope names the are returned are e.g. 1 meter, 2 meter, etc...
```
jnation3406, line 318: that instead of the?

observation_portal/proposals/models.py
```diff
@@ -10,7 +10,9 @@
 from collections import namedtuple
 import logging
 
+
```
jnation3406, line 13: is this second blank line intentional?

jnation3406 (approved):

observation_portal/proposals/models.py
eheinrich: Nope, removed it

observation_portal/common/configdb.py
eheinrich: Done

PR #167, opened 2020-12-15
Include proper motion for all example targets

timbeccue:
Adding the proper motion fields to all example targets (default value of 0) avoids errors with the scheduler running in the ocs_example setup.


eheinrich (approved):
Looks good

PR #168, opened 2020-12-18
Fix/integration testing fixes

eheinrich:
These are some fixes that I made after a round of final testing of the new frontend.
- Updated some settings used by `django-storages` to fix genenerating s3 urls. The sciapps pages previously used a custom templatetag to generate the presigned s3 url for a user to access, but actually `django-storages` generates one for you automatically when the `url` prop of a `FileField` is accessed, as long as the settings are set up correctly. I couldn't see why we needed to set our own custom domain (the `AWS_S3_CUSTOM_DOMAIN` setting) and with it removed, the presigned url gets generated correctly and uploads work. But let me know if you remember why we needed that setting.
- Fixed a validation issue with sciapp submissions to allow a user to submit a sciapp draft that they had previously saved that already has a PDF associated with it, without requiring them to include that same PDF in the submission request body. I also added a check to prevent a user from clearing the PDF that they had previously uploaded when submitting a draft, if that sciapp requires a PDF.

### DIFF
--- a/observation_portal/settings.py
+++ b/observation_portal/settings.py
@@ -165,7 +165,7 @@ OAUTH2_PROVIDER = {
 }
 
 CORS_ORIGIN_ALLOW_ALL = True
-CORS_URLS_REGEX = r'^/(api|accounts)/.*$|^/o/.*'
+CORS_URLS_REGEX = r'^/(api|accounts|media)/.*$|^/o/.*'
 LOGIN_REDIRECT_URL = '/accounts/loggedinstate/'
 
 # Internationalization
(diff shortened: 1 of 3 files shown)


observation_portal/settings.py
```diff
@@ -187,21 +187,21 @@
 # https://docs.djangoproject.com/en/2.1/howto/static-files/
 
 AWS_STORAGE_BUCKET_NAME = os.getenv('AWS_BUCKET_NAME', 'observation-portal-test-bucket')
-AWS_REGION = os.getenv('AWS_REGION', 'us-west-2')
+AWS_S3_REGION_NAME = os.getenv('AWS_REGION', 'us-west-2')
 AWS_ACCESS_KEY_ID = os.getenv('AWS_ACCESS_KEY_ID', None)
 AWS_SECRET_ACCESS_KEY = os.getenv('AWS_SECRET_ACCESS_KEY', None)
-AWS_S3_CUSTOM_DOMAIN = 's3-us-west-2.amazonaws.com/{}'.format(AWS_STORAGE_BUCKET_NAME)
```
jnation3406, line 193: yeah I'm not sure why we had this... maybe just our lack of knowledge back then. If it works with uploading and accessing older uploaded files, then that's fine

observation_portal/settings.py
```diff
@@ -187,21 +187,21 @@
...
+AWS_S3_SIGNATURE_VERSION = 's3v4'
 
 STATICFILES_DIRS = [os.path.join(BASE_DIR, 'static')]
 STATIC_URL = '/static/'
 STATIC_ROOT = '/static/'
 STATICFILES_DIRS = [os.path.join(BASE_DIR, 'static')]
 STATICFILES_STORAGE = os.getenv('STATIC_STORAGE', 'django.contrib.staticfiles.storage.StaticFilesStorage')
 MEDIAFILES_DIR = os.getenv('MEDIAFILES_DIR', 'media')
-MEDIA_URL = "https://{}/{}/".format(AWS_S3_CUSTOM_DOMAIN, MEDIAFILES_DIR) if AWS_ACCESS_KEY_ID else '/media/'
+MEDIA_URL = '' if AWS_ACCESS_KEY_ID else '/media/'
```
jnation3406, line 204: I'm guessing we just don't even use MEDIA_URL anymore since we don't use templates? Previously we used this to determine if it should use AWS for media files, or local media file hosting for dev - I'm guessing the local hosting doesn't work anymore in the frontend since it's probably just pointing to s3 now?

jnation3406 (approved):

observation_portal/settings.py
eheinrich: No actually it still works. If you don't set the `MEDIA_STORAGE` environment variable to use the s3 backend, then local file storage will be used by default, and if you have `DEBUG=True`, then the app itself will serve media files at `/media/`, similar to how we handle static files during development.

PR #171, opened 2021-01-16
Add NOT_ATTEMPTED state for Observations and Configuration Statuses

jnation3406:
This is for failing observations that won't have enough time to finish before they start, and for "failing" ones that are in the past and were never attempted.

As discussed with Mark, the rules are if all configuration statuses are NOT_ATTEMPTED, the observation is NOT_ATTEMPTED.
If any configuration status is NOT_ATTEMPTED and all others are PENDING, the observation is still PENDING (we don't know yet if it will be NOT_ATTEMPTED, or FAILED, need the rest of them to be updated first). If any configuration status is NOT_ATTEMPTED and any other is not PENDING, then the observation is FAILED. 

I think I'll also need a small change to the frontend for coloring NOT_ATTEMPTED observations differently.


markBowman (commented):
Looks good as far as I can tell

eheinrich (approved):
Looks good to me!

PR #172, opened 2021-01-19
Fix/separate frontend deploy fixes

eheinrich:
A couple of minor fixes from deploying the separate frontend:
- Needed to add a configurable `CSRF_TRUSTED_HOSTS` to allow the forms that are passed through to the frontend to work. I hadn't caught this before since the host configured in the ingress of the dev deploy and the hostname of the dev deploy pod were both the same, so there was no issue. However, with the prod deploy, the ingress host is different from the pod hostname so this is needed to let the frontend submit a form.
- Added a div wrapping the content of the base HTML template. This template is used for the various registration forms, and the library that I am using in the frontend to extract/ display the form in the frontend will only be able to extract the form if the element is not at the top level.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,6 +9,7 @@ media/
 .tox/
 .git/
 .venv/
+.env/
 **/*.swp
 **/*.pyc
 **/__pycache__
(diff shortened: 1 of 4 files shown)


jnation3406 (approved):

eheinrich:
Ran the tests locally, they all passed.

PR #173, opened 2021-01-20
Validate that the defocus value, when passed in, is a number

eheinrich:
Check that the value for defocus that is passed in is a number. We ultimately want to add a validation schema to configdb under instrument types that can then be used to validate the extra_params of instrument_configs, and then use that to validate all the values passed in. However, this is just to make sure that a user does not pass in an invalid value in the meantime, until we start on the stories to further generalize the portal.


markBowman (approved):
Thank you, thank you, thank you :-)

eheinrich:
I ran the tests locally, they passed.

jnation3406 (approved):

PR #175, opened 2021-01-26
Feature/gh actions

eheinrich:
Switch from using TravisCI for running tests and publishing docker images to using GitHub Actions instead, seeing as Travis has limited their support for open source projects.

The behavior should be the same as before:
- when making a new commit, tests are run
- If a new tag is created, tests are run, and also a docker image is built and pushed with the image tag set to the git tag.

I made a test tag to confirm that a docker image is built and ends up in the docker hub.

The [`GITHUB_TOKEN`](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret) secret is a token that is generated at the start of a run and that can be used to authenticate a run. It then expires at the end of each run. The `GITHUB_TOKEN` is allowed to be [read/used in workflows triggered from forks](https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow), but all other secrets are not. For this workflow, this means that coverage will still be checked on forked repository, but the build and push image step will only work in a non-forked repository.


.github/workflows/build.yaml
```diff
@@ -0,0 +1,80 @@
...
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
jnation3406, line 73: where does it get the secrets from?

jnation3406 (approved):
awesome! goodbye travis

dmcollom (approved):

.github/workflows/build.yaml
eheinrich: There are a few different places secrets can be stored, but I've placed these in the secrets at the organization level which you can get to if you go to the organization settings, and then click on the "Secrets" tab.

PR #176, opened 2021-01-26
add RETRY_LIMIT  as request state and set it when that request has nu…

jnation3406:
…mber of observations fail beyond the limit

If you think the state should be called something other then RETRY_LIMIT I can change it. I picked it over MAXIMUM_RETRIES and I wanted to keep it short but meaningful, but we could do something adding another word like MAXIMUM_RETRIES_REACHED or RETRY_LIMIT_EXCEEDED if you think it needs more description.


observation_portal/common/state_changes.py
```diff
@@ -273,9 +284,9 @@ def aggregate_request_states(request_group):
     """Aggregate the state of the request group from all of its child request states"""
     request_states = [request.state for request in Request.objects.filter(request_group=request_group)]
     # Set the priority ordering - assume AND by default
-    state_priority = ['WINDOW_EXPIRED', 'PENDING', 'COMPLETED', 'CANCELED']
+    state_priority = ['WINDOW_EXPIRED', 'PENDING', 'COMPLETED', 'RETRY_LIMIT', 'CANCELED']
     if request_group.operator == 'MANY':
-        state_priority = ['PENDING', 'COMPLETED', 'WINDOW_EXPIRED', 'CANCELED']
+        state_priority = ['PENDING', 'COMPLETED', 'WINDOW_EXPIRED', 'RETRY_LIMIT', 'CANCELED']
```
jnation3406, line 289: Aggregating Request states into RequestGroup states is a little strange since it's not clear how what RequestGroup state corresponds to having Requests in a variety of states. This describes the order in which we decide the state, based on if any of the Requests are in that state. In this case I put RETRY_LIMIT behind WINDOW_EXPIRED, but I would be open to putting it in front of that instead if you think RETRY_LIMIT is a more important top level state then WINDOW_EXPIRED...

observation_portal/common/state_changes.py
```diff
@@ -225,15 +230,21 @@ def modify_ipp_time_from_requests(ipp_val, requests_list, modification='debit'):
...
+    acceptability_threshold = request.acceptability_threshold
     observation_state = get_observation_state(configuration_statuses)
     completion_percent = exposure_completion_percentage(configuration_statuses)
     if isclose(acceptability_threshold, completion_percent) or \
             completion_percent >= acceptability_threshold or \
             observation_state == 'COMPLETED':
         return 'COMPLETED'
-    return request_state
+    # If a nonzero MAX_RETRIES_PER_REQUEST is set and the observation failed, check that condition
+    if settings.MAX_RETRIES_PER_REQUEST and observation_state == 'FAILED':
```
jnation3406, line 243: I'm only counting FAILED observations for the retry limit, because I think FAILED is the only state that occurs due to the observation guiding/acquisition type issues we are trying to avoid here. ABORTED should only occur when it's forcefully cancelled by the scheduler for another observation, so that shouldn't count towards the limit.

observation_portal/proposals/notifications.py
```diff
@@ -14,6 +14,29 @@ def users_to_notify(requestgroup):
     ]
 
 
+def request_notifications(request):
```
markBowman, line 17: We should probably run this email template past Nikolaus for his approval.

markBowman (approved):
Nice work. Thank you.

observation_portal/common/state_changes.py
eheinrich: Sounds good to me

observation_portal/common/state_changes.py
eheinrich: I think this is reasonable. I honestly can't think of a strong case that would require one to go before the other.

observation_portal/common/test_state_changes.py
```diff
@@ -394,12 +396,55 @@ def test_configuration_statuses_not_complete_or_failed_use_initial(self):
...
+            dmixer.blend(Summary, configuration_status=cs, time_completed=0)
+        observation2.refresh_from_db()
+        self.assertEqual(observation2.state, 'FAILED')
+        # so request should enter the RETRY_LIMIT state since it has failed "max retries" times
+        request_state = get_request_state_from_configuration_statuses(
+            initial_state, self.request, observation2.configuration_statuses.all()
+        )
+        self.assertEqual(request_state, 'RETRY_LIMIT')
+        # Test that an email is sent out for the Request
+        self.assertEqual(len(mail.outbox), 1)
```
eheinrich, line 443: You should wait for the dramatiq worker to finish before checking what's in the mailbox: https://github.com/Bogdanp/django_dramatiq#using-unittest

eheinrich (approved):
Looks good! I do prefer something more along the lines of your other suggestions, either `MAXIMUM_RETRIES_REACHED` or `RETRY_LIMIT_EXCEEDED` since `RETRY_LIMIT` sounds more like a number instead of a state.

I do have a thought about using the word "retries". The setting is `MAX_RETRIES_PER_REQUEST`, but to me that sounds like it's how many times we should try to observe the request _again_, instead of the number of attempts we should make to observe the request.

What about using the word "attempt" instead - so, `MAX_ATTEMPTS_PER_REQUEST` for the setting, and `MAX_ATTEMPTS_REACHED` for the state.

observation_portal/proposals/notifications.py
jnation3406: Should have mentioned in the review that I worked with him for the email subject and message

PR #178, opened 2021-02-05
Multiple target soar

jnation3406:
Like we discussed, this is doing a simple intersection of rise-sets for each target to get total request rise-set windows. It then uses that and averages the airmass values of all targets and constraints within the rise-set windows. I hate adding the SOAR specific code in here, but we will clean it up soon with the stories we have planned.

### DIFF
--- a/observation_portal/requestgroups/test/test_request_utils.py
+++ b/observation_portal/requestgroups/test/test_request_utils.py
@@ -324,7 +324,7 @@ class TestMultipleTargetRequestIntervals(BaseSetupRequest):
         # )
         # mixer.blend(Constraints, configuration=self.configuration2)
 
-    def test_request_intervals_for_one_week(self):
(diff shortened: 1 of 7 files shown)


observation_portal/requestgroups/test/test_request_utils.py
```diff
@@ -302,6 +305,98 @@ def test_request_intervals_for_one_week_overlapping_downtime(self, downtime_data
         self.assertEqual(intervals, truth_intervals)
 
 
+class TestMultipleTargetRequestIntervals(BaseSetupRequest):
+    def setUp(self):
+        super().setUp()
+        # Set up a second configuration with a different Target
```
eheinrich, line 311: Commented out code

eheinrich (approved):
Looks good to me! Yeah we'll get rid of the SOAR specific code soon.

PR #179, opened 2021-02-09
Fix/multiple timeallocations

eheinrich:
Changes so that multiple `TimeAllocation` objects with the same instrument_type, proposal, and semester are not allowed to be created in the database. This is something that we already assume to be true in various places in the code, and allowing duplicates does not make sense.

- Added a uniqueness constraint to the `TimeAllocation` model requiring time allocations to be unique for the same combination of semester, instrument_type, and proposal. This will keep an admin user on the proposal change admin page from saving time allocations that violate this constraint, and an `IntegrityError` will be raised if an attempt is made using the ORM to save a violating allocation.
- Added some validation into the science applications app keeping users from inputting time requests that could be turned into duplicate time allocations. If users were allowed to submit duplicate time requests, the proposal admin would have to decide which time request is valid at time of porting, which they would probably have to reach out to the submitter of the science application to figure out. Might as well require the user to sort it out when they are constructing their application. I considered adding a database constraint for the `TimeRequest` model as well, but whereas we definitely do not want to allow duplicate `TimeAllocation` objects, and keeping users from submitting duplicate time requests seems like a good idea, I think it's a softer requirement and is potentially OK to break. So I just went with validating at the app level, but let me know if you disagree.


jnation3406 (approved):
I guess I'm not sure the value of allowing multiple TimeRequests for an application, since that forces the admin to then reach out and clarify which TimeRequest to approve/use... It seems like just not allowing an application to have multiple for the same instrument/semester would save that step, but if it's much harder to do or there are other issues, this is fine since it's already an edge case.

eheinrich:
Right, sorry if I wasn't clear in my initial post - I did indeed add validation into the scienceapplications serializer to keep a user from submitting multiple time requests for the same instrument/semester for their science application. What I did not do was take it a step further and add a uniqueness constraint into the database for the `TimeRequest` model, since validating in the serializer seemed sufficient.

jnation3406:
oh okay, yeah I misunderstood. This seems fine then!

PR #181, opened 2021-02-18
Add InstrumentType Cerberus Validation

mgdaily:
This PR adds the ability to validate configs from `validation_schema` located inside of an `instrument_type` in ConfigDB.

* Refactors some common validation code into the `ValidationHelper` class. `InstrumentTypeValidationHelper` and `ModeValidationHelper` both inherit from this class. I thought the best way to do this was to define a loose interface for a generic `ValidationHelper`, so I have left the constructor and `validate` methods as abstract. I moved the`cerberus_validation_error_to_str` helper function into the class as a private method since it seemed to logically fit there, and defined a generic `validate_config` private method as well. Everything else, including custom error handling/messaging is left up to the subclasses.
* Removes specific defocus check from ConfigurationSerializer
* Adds some unit tests for the ValidationHelper classes. There are integration tests (updated too), but the story Jon wrote included a task to add unit tests. Let me know if `test_api.py` is an appropriate place to put them.

ConfigDB PR: https://github.com/observatorycontrolsystem/configdb/pull/19


observation_portal/requestgroups/serializers.py
```diff
@@ -358,6 +401,10 @@ def validate(self, data):
                     raise serializers.ValidationError(_(str(cdbe)))
             readout_validation_helper = ModeValidationHelper('readout', instrument_type, modes['readout'])
             instrument_config = readout_validation_helper.validate(instrument_config)
+
+            validator = InstrumentTypeValidationHelper(instrument_type)
```
jnation3406, line 405: maybe name this instrument_validator or general_instrument_validator or something

jnation3406 (approved):

observation_portal/requestgroups/serializers.py
mgdaily: Good catch, I missed that pattern

observation_portal/requestgroups/serializers.py
```diff
@@ -24,42 +25,92 @@
...
 
-def cerberus_validation_error_to_str(validation_errors: dict) -> str:
-    error_str = ''
-    for field, value in validation_errors.items():
-        if (isinstance(value, list) and len(value) == 1 and isinstance(value[0], dict)):
-            error_str += f'{field}{{{cerberus_validation_error_to_str(value[0])}}}'
-        else:
-            error_str += f'{field} error: {", ".join(value)}, '
+class ValidationHelper(ABC):
+    """Base class for validating configurations"""
```
eheinrich, line 37: The word "configurations" here means some dictionary with fields to validate, not necessarily a configuration in terms of the Configurations of a Request?

eheinrich (approved):
Looks good! I just have a minor question about a docstring. I think the unit tests would fit better in either `test.py` or in `test_helpers.py`.

observation_portal/requestgroups/serializers.py
mgdaily: Good point. It's not necessarily a Configuration from a Request. Presumably it could be any structure that needs to be validated via Cerberus.

Cerberus' docs use the term "document" a lot -  which seems more appropriate?

eheinrich: Yeah I think "document" sounds good! Definitely would make it more clear.

PR #183, opened 2021-02-22
Base project

jnation3406:
The project has been modified to build and deploy a library called `django-ocs-observation-portal` to pypi. The serializers have been modified to load a class via dotpath specified from the settings.py / environment variables. Even though this project it a library, it contains the 'base' version of the settings.py that is expected to be included and overridden from the project using this library. The base settings.py also allows us to still run the unit tests on deploy of the library, and create a usable base docker image.

The forkable project using this library is here: https://github.com/observatorycontrolsystem/observation-portal-project/pull/1


setup.py
```diff
@@ -0,0 +1,62 @@
...
+        'License :: OSI Approved :: BSD License',
+        'Operating System :: OS Independent',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.7',
+        'Topic :: Scientific/Engineering :: Astronomy',
+        'Topic :: Scientific/Engineering :: Physics'
+    ],
+    keywords=['observations', 'astronomy', 'astrophysics', 'cosmology', 'science', 'ocs', 'observatory'],
+    packages=find_packages(),
+    use_scm_version=True,
```
eheinrich, line 28: Oh cool, so does using `setuptools_scm` mean that the version of the package uploaded to pypi is the git tag?

eheinrich (approved):
Nice! How were you thinking the customizable serializer paths should be documented?

setup.py
jnation3406: I believe so, but I haven't run the pipeline on a tag yet.

jnation3406:
I have all the serializer paths commented out in the project repo's settings.py to uncomment and change to override. Otherwise I just point them to this library settings.py file from that projects readme to look up. I guess since I made them overridable by environment variable, we should list them in the huge environment variable list in the readme of the library? Or is there a better place to put them?

eheinrich:
> I have all the serializer paths commented out in the project repo's settings.py to uncomment and change to override. Otherwise I just point them to this library settings.py file from that projects readme to look up. I guess since I made them overridable by environment variable, we should list them in the huge environment variable list in the readme of the library? Or is there a better place to put them?

Yeah I think adding them to the list of environment variables sounds good.

PR #187, opened 2021-03-13
Configdb changes

jnation3406:
Changes in configdb this accounts for:
1. front_padding -> observation_front_padding
2. PER_CONFIGURATION_STARTUP_TIME constant -> configuration_front_padding in instrument_type
3. instrument_category added to instrument_type for IMAGE/SPECTRO from instruments api
4. configuration_type is a dictionary now, containing a change overhead, and whether the type is schedulable or not
5. Generic modes now have a schedulable field as well

Two slight differences in calculating durations: 
1. For REPEAT_EXPOSE configurations, the repeat_duration now implictely includes the configuration_front_padding (before it did not, and that amount was added in later).
2. The config_change_overhead is now per configuration type rather than per instrument_type. Therefore it is only applied when the configuration type changes, and the type it is changing too has a non-zero config_change_overhead. This slightly effects the time estimated for observations that use this, which for is us just Floyds ARC/LAMP_FLAT.


observation_portal/requestgroups/duration_utils.py
```diff
@@ -58,19 +58,21 @@ def get_configuration_duration(configuration_dict):
...
         req_info = {'duration': get_request_duration(req)}
-        conf_durations = [get_configuration_duration(conf) for conf in req['configurations']]
+        conf_durations = []
+        for conf in req['configurations']:
+            request_overheads = configdb.get_request_overheads(conf['instrument_type'])
+            conf_durations.append(get_configuration_duration(conf, request_overheads))
         req_info['configurations'] = conf_durations
         rise_set_intervals = get_filtered_rise_set_intervals_by_site(req, is_staff=is_staff)
         req_info['largest_interval'] = get_largest_interval(rise_set_intervals).total_seconds()
-        req_info['largest_interval'] -= (PER_CONFIGURATION_STARTUP_TIME)
```
jnation3406, line 73: This "largest_interval" is used for computing repeat_expose / fill_window stuff. By removing this, we are making the largest_interval larger and implicitly including the per configuration_front_padding into the repeat_duration that will get set, rather than adding it later.

observation_portal/requestgroups/duration_utils.py
```diff
@@ -58,19 +58,21 @@ def get_configuration_duration(configuration_dict):
         conf_duration['duration'] = configuration_dict['repeat_duration']
     else:
         conf_duration['duration'] = sum([icd['duration'] for icd in instrumentconf_durations])
-    conf_duration['duration'] += (PER_CONFIGURATION_STARTUP_TIME)
+        conf_duration['duration'] += request_overheads['config_front_padding']
```
jnation3406, line 61: Here we only apply the front_padding to non-repeat-expose configurations, since the repeat_duration is now expected to contain the config_front_padding within it.

observation_portal/requestgroups/duration_utils.py
```diff
@@ -194,10 +197,10 @@ def get_complete_configurations_duration(configurations_list, start_time, priori
...
                     duration += request_overheads['guiding_overheads'][configuration_dict['guiding_config']['mode']]
 
-            # TODO: find out if we need to have a configuration type change time for spectrographs?
-            if configdb.is_spectrograph(configuration_dict['instrument_type']):
-                if previous_conf_type != configuration_dict['type']:
-                    duration += request_overheads['config_change_overhead']
+            # Certain Configuration Types for certain Instrument Types will have a non-zero config_change_overhead.
+            # For instance, this could account for Lamp startup times when first switching to an ARC or LAMP_FLAT configuration.
+            if previous_conf_type != configuration_dict['type']:
+                duration += configuration_types[configuration_dict['type']]['config_change_overhead']
```
jnation3406, line 203: Removed the check for if it is a spectrograph. Instead any configuration_type that is switched to will have its config_change_overhead added, but for most types that will be zero. If you only have a single configuration_type and it has a non-zero config_change_overhead, then it will not be applied since you aren't changing types, but I think that can get covered by the front_padding. If you think not, we could change this logic to apply the config_change_overhead for the "first" configuration too...

observation_portal/requestgroups/serializers.py
```diff
@@ -153,13 +152,15 @@ def validate(self, config_dict) -> dict:
             ))
         return validated_config_dict
 
-    def mode_exists(self, mode_value: str) -> bool:
+    def mode_exists_and_is_schedulable(self, mode_value: str) -> bool:
```
jnation3406, line 155: Check if both this mode exists, and if it does that it is schedulable. The error is the same either way if not, that the mode is not available.

observation_portal/requestgroups/test/test_api.py
```diff
@@ -1796,7 +1808,7 @@ def test_fill_window_confined_window_fills_the_window(self):
         rg = response.json()
         self.assertGreater(rg['requests'][0]['configurations'][0]['repeat_duration'],
                            initial_repeat_duration)
-        self.assertEqual(rg['requests'][0]['configurations'][0]['repeat_duration'], 430.0)
+        self.assertEqual(rg['requests'][0]['configurations'][0]['repeat_duration'], 446.0)
```
jnation3406, line 1811: repeat_duration now includes the configuration_front_padding, which is 16s

observation_portal/requestgroups/duration_utils.py
eheinrich: Wait, doesn't the `config_change_overhead` always apply to the first configuration with this logic? On the first pass through the loop, `previous_conf_type` is an empty string so won't match any configuration. I do think it makes sense to apply the change overhead to the first configuration, it seems like that would allow us to model the overheads better. But were you thinking it would just be better to capture it with the front padding?

observation_portal/requestgroups/duration_utils.py
eheinrich: I think that sounds good

eheinrich (approved):
I think it looks good! Just one question.

observation_portal/requestgroups/duration_utils.py
jnation3406: you are correct, I had assumed we pre-populated with the first configuration type but we don't. It is probably better this way, since it should more accurately model times...

PR #190, opened 2021-04-02
Change the telescope_class filter on schedulable_requests to be teles…

jnation3406:
…cope_class and take multiple input params

This change is to support scheduling on more than one telescope_class but not all with the scheduler.

### DIFF
--- a/observation_portal/requestgroups/viewsets.py
+++ b/observation_portal/requestgroups/viewsets.py
@@ -77,9 +77,7 @@ class RequestGroupViewSet(ListAsDictMixin, viewsets.ModelViewSet):
         current_semester = Semester.current_semesters().first()
         start = parse(request.query_params.get('start', str(current_semester.start))).replace(tzinfo=timezone.utc)
         end = parse(request.query_params.get('end', str(current_semester.end))).replace(tzinfo=timezone.utc)
-        telescope_classes = request.query_params.get('telescope_classes')
-        if telescope_classes:
-            telescope_classes = telescope_classes.split(',')
+        telescope_classes = request.query_params.getlist('telescope_class')
         # Schedulable requests are not in a terminal state, are part of an active proposal,
         # and have a window within this semester
         instrument_config_query = InstrumentConfig.objects.prefetch_related('rois')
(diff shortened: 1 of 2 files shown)


observation_portal/requestgroups/viewsets.py
```diff
@@ -77,7 +77,9 @@ def schedulable_requests(self, request):
         current_semester = Semester.current_semesters().first()
         start = parse(request.query_params.get('start', str(current_semester.start))).replace(tzinfo=timezone.utc)
         end = parse(request.query_params.get('end', str(current_semester.end))).replace(tzinfo=timezone.utc)
-        telescope_class = request.query_params.get('telescope_class')
+        telescope_classes = request.query_params.get('telescope_classes')
```
eheinrich, line 80: To be more in line how our other endpoints that allow multiple values for certain query params work, what do you think about letting a user set the `telescope_class` param one or more times (so, for example, `?telescope_class=0m4&telescope_class=1m0`), and then from here get the list of telescope classes using the [QueryDict's getlist method](https://docs.djangoproject.com/en/3.1/ref/request-response/#querydict-objects)?

jnation3406: didn't know about the getlist method on querydicts but that does sound better. I'll see about switching it over

observation_portal/requestgroups/viewsets.py
```diff
@@ -101,8 +103,8 @@ def schedulable_requests(self, request):
             Prefetch('proposal', queryset=Proposal.objects.only('id').all()),
             Prefetch('submitter', queryset=User.objects.only('username', 'is_staff').all())
         ).distinct()
-        if telescope_class:
-            queryset = queryset.filter(requests__location__telescope_class__iexact=telescope_class)
+        if telescope_classes:
+            queryset = queryset.filter(requests__location__telescope_class__in=telescope_classes)
```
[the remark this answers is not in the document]
jnation3406: it doesn't, but I think we use lowercase everywhere for telescope class definitions. I guess it couldn't hurt to convert things to lowercase to be safe...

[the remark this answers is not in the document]
jnation3406: actually on second though, not doing case-insensitive is probably more appropriate here anyway. That way users must specify the proper case as is used within configdb.

observation_portal/requestgroups/viewsets.py
jnation3406: okay, changed the keyword to `telescope_class` using getlist for multiples

eheinrich (approved):

PR #192, opened 2021-04-09
update downtimedb to support applying downtimes to specific instrumen…

jnation3406:
…t_types or all instruments on the telescope

Changes the downtime structure to be a dict ordered by resource and then by instrument_type or 'all' if it should apply to all instrument_types on that resource. Added a few unit tests to test out these cases.


observation_portal/common/downtimedb.py
```diff
@@ -34,35 +34,39 @@ def _get_downtime_data():
...
             start = datetime.strptime(interval['start'], DOWNTIME_DATE_FORMAT).replace(tzinfo=timezone.utc)
             end = datetime.strptime(interval['end'], DOWNTIME_DATE_FORMAT).replace(tzinfo=timezone.utc)
-            downtime_intervals[resource].append({'type': 'start', 'time': start})
-            downtime_intervals[resource].append({'type': 'end', 'time': end})
+            downtime_intervals[resource][instrument_type].append({'type': 'start', 'time': start})
+            downtime_intervals[resource][instrument_type].append({'type': 'end', 'time': end})
 
         for resource in downtime_intervals:
-            downtime_intervals[resource] = Intervals(downtime_intervals[resource])
+            for instrument_type, intervals in downtime_intervals[resource].items():
```
eheinrich, line 54: Why loop through `.items()` like this? `intervals` is never used.

eheinrich (approved):
Looks good

observation_portal/common/downtimedb.py
jnation3406: oops i should be using intervals on the next line

PR #193, opened 2021-04-13
prevent timeallocations in use from being changed or deleted

jnation3406:
This prevents you from deleting a TimeAllocation in the admin interface if a Request exists that uses that TA. It also prevents you from changing the instrument_type or semester on a time allocation that is in use.


observation_portal/proposals/tests.py
```diff
@@ -295,3 +310,15 @@ def test_activate_selected(self):
         )
         for proposal in self.proposals:
             self.assertEqual(Proposal.objects.get(pk=proposal.id).active, True)
+
+    def test_clean_time_allocation(self):
+        ta_dict = self.time_allocation.as_dict()
+        taf = TimeAllocationForm(data=ta_dict, instance=self.time_allocation)
+        self.assertTrue(taf.is_valid())
+
+        # Now attempt to change the semester and see it fail
```
jnation3406, line 319: I tried to add this unit test. It works in isolation, but if I run it uncommented with all other unit tests, it breaks _other_ unrelated unit tests! I've traced it to the TimeAllocationForm with data with the changed semester id, but have no idea why this break other tests?

eheinrich: Interesting, nothing is standing out to me about the test that would break other tests.

jnation3406: yeah... it is pretty frustrating since it doesn't really have any clear reason why it would impact other tests. Its basically just calling form validation which shouldn't make any changes

observation_portal/proposals/forms.py
```diff
@@ -26,3 +27,33 @@ class Meta:
...
+
+    def clean(self):
+        cleaned_data = super().clean()
+        if self.instance.pk is None:
+            return
+        instrument_type_changed = cleaned_data.get('instrument_type') != self.instance.instrument_type
+        semester_changed = cleaned_data.get('semester') != self.instance.semester
+        if instrument_type_changed or semester_changed:
+            # instrument_type has changed. We should block this if the old instrument_type was in use
+            requestgroups = RequestGroup.objects.filter(proposal=self.instance.proposal)
```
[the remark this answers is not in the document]
jnation3406: yeah I guess so, I wasn't worried about optimization since this is just in an admin action, so pretty rare and non time critical

PR #194, opened 2021-04-15
Per instrumnt change time

jnation3406:
I don't know what I was thinking when I made this branch name. It should be per telescope class change time, not per instrumnt. 

I added caching the last_change time per telescope_class, but also kept a generic time of "all" to quickly get the overall last change time in cases that need that. The endpoint to get the last change time takes zero or more telescope_classes as well, and returns the latest of all in the case of multiple telescope classes, or returns the "all" option (which is also the latest overall) if no telescope_class is set


observation_portal/requestgroups/serializers.py
```diff
@@ -758,7 +760,7 @@ def create(self, validated_data):
             'tracking_num': request_group.id,
             'name': request_group.name
         }})
-        cache.set('observation_portal_last_change_time', timezone.now(), None)
+        cache.set('observation_portal_last_change_time_all', timezone.now(), None)
```
eheinrich, line 763: This probably doesn't matter since it would more often than not be in the microseconds, but do you see any issues with the "all" last change time being slightly different from the most recent telescope_class last change time since `timezone.now()` is called for every call to `cache.set(...)`?

eheinrich (approved):

observation_portal/requestgroups/serializers.py
jnation3406: i think microseconds don't matter (since we just use this to determine if we need to re-schedule yet or not), but I could change it to use the same now...

observation_portal/requestgroups/test/test_api.py
```diff
@@ -2825,11 +2830,47 @@ def test_last_change_date_is_updated_when_request_is_submitted(self):
...
 
+    def test_last_change_date_telescope_class_is_updated_when_request_is_submitted(self):
+        last_change_cached = self.locmem_cache.get('observation_portal_last_change_time_1m0')
+        self.assertIsNone(last_change_cached)
+        rg = self.generic_payload.copy()
+        response = self.client.post(reverse('api:request_groups-list'), data=self.generic_payload)
+
+        response = self.client.get(reverse('api:last_changed') + '?telescope_class=1m0')
+        last_change = response.json()['last_change_time']
+        self.assertAlmostEqual(datetime_parser(last_change), timezone.now(), delta=timedelta(minutes=1))
```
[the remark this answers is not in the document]
jnation3406: I guess saving the time before posting the request and then using an assertGreater with that would work too. This just uses assertAlmostEqual within a minute because timezone.now() will obviously be a little ahead, but just microseconds (Assuming posting the request just updated the time), so it will be almost equal.

observation_portal/requestgroups/serializers.py
eheinrich: Yeah.. the reason I lean towards changing it is mostly cause having the times be exactly the same is what the user would expect, so if there's a bit of a difference it might be momentarily confusing. The test `test_last_change_date_telescope_class_is_updated_when_request_is_submitted` even checks that those times are equal. Which pass of course because it uses mocked time, but it's still slightly different from the behavior of the running app. I'll leave it up to you.

observation_portal/requestgroups/test/test_api.py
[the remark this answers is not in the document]
eheinrich: Gotcha. I think comparing to a time right before posting the request would be a bit more clear in the intention of the test.

observation_portal/requestgroups/serializers.py
jnation3406: Okay, I've changed all of them to use the same now value for _all and per telescope caches. I agree this seems fine and means there will be less discrepancies

PR #196, opened 2021-04-21
two potential fixes for the failing time allocation deletion unit test

jnation3406:
Either the line at the end of the unit test (nulling out the global semesters), or the change from a global variable to using the locmem cache will fix this. Which do you prefer? I can't remember why exactly we made this a global variable instead of something from the cache in the first place, but to me it seems like local memory cache would work best. I think this was an optimization to "cache" the semesters during each instance of processing a request, because we would query them many times through the course of validating a request I think... So a locmem cache should work just as well in that case.


eheinrich:
I agree, I prefer using the locmem cache to store the semesters. I don't remember why semesters were saved in a global variable either, but your guess sounds reasonable to me.

jnation3406:
okay, I've removed the test workaround and left the locmem caching. Added another test case as well.

eheinrich (approved):

PR #197, opened 2021-04-24
Link the sciapp detail url setting from the admin page

eheinrich:
These changes to the links are to make proposal management easier.
- The preview link column in the sciapplications list page on the admin site now optionally includes a link to the value of the `SCIENCE_APPLICATION_DETAIL_URL` setting if it is set.
- Also if the `SCIENCE_APPLICATION_DETAIL_URL` setting is set, use that url as the destination for the "View on Site" button on the change page of a sciapplication in the admin site.

### DIFF
--- a/observation_portal/sciapplications/admin.py
+++ b/observation_portal/sciapplications/admin.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 from django.contrib import admin
-from django.utils.html import format_html
+from django.utils.html import format_html_join
 from django.urls import reverse
-
+from django.conf import settings
 
 from .models import (
     Instrument, Call, ScienceApplication, TimeRequest, CoInvestigator,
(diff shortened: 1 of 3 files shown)


jnation3406 (approved):
This seems fine... I wonder if maybe just providing a link in the admin page to the combined pdf would be enough, since it seems like that is what Nikolaus was looking for. Then it wouldn't need to link to the frontend.

eheinrich:
Actually the combined PDF is generated by the frontend not the backend, and it is linked from the science application detail page. Do you think we need a separate link directly to the combined PDF?

jnation3406:
> Actually the combined PDF is generated by the frontend not the backend, and it is linked from the science application detail page. Do you think we need a separate link directly to the combined PDF?

Ah i forgot its in the frontend now. No this is fine then, I was just trying to see if maybe there was an easier way to get Nikolaus what I think he wanted

PR #204, opened 2021-06-15
Education user registration to basic mode

zemogle:
Change registration behaviour so there are no hidden presets or the concept of an education user. I've also fixed the bug in "Pricipal Investigator" > "Principal Investigator"


zemogle:
It seems I was using the wrong starting code base. Just to be sure, I'll create a new PR

PR #205, opened 2021-06-16
Update basic mode

zemogle:
Basic mode on registration page plus test coverage. Also fixed a typo in "Principal Investigator" model


observation_portal/accounts/forms.py
```diff
@@ -32,11 +33,7 @@ def save(self, commit=True):
             user=new_user_instance,
             title=self.cleaned_data['title'],
             institution=self.cleaned_data['institution'],
-            education_user=self.cleaned_data['education_user'],
```
eheinrich, line 35: Is the intention to manually update who is an education user from now on?

observation_portal/proposals/models.py
```diff
@@ -232,7 +232,7 @@ class Membership(models.Model):
     PI = 'PI'
     CI = 'CI'
     ROLE_CHOICES = (
-        (PI, 'Pricipal Investigator'),
```
eheinrich, line 235: Can you create a new migration for this change?

observation_portal/accounts/forms.py
```diff
@@ -12,18 +12,19 @@ class CustomRegistrationForm(RegistrationFormTermsOfService, RegistrationFormUni
     last_name = forms.CharField(max_length=200)
     institution = forms.CharField(max_length=200)
     title = forms.CharField(max_length=200)
-    education_user = forms.BooleanField(required=False)
+    simple_interface = forms.BooleanField(label="Use Basic Mode", help_text="Hide advanced fields when making an observation request.", required=False)
```
eheinrich, line 15: The same help text is also written on line 27. Does it need to be set in these two different places?

eheinrich (commented):
Just a few questions

observation_portal/proposals/models.py
zemogle: Done 👍

observation_portal/accounts/forms.py
zemogle: I've removed the help_text in Meta as that version is not being used (there is help_text for `username` which doesn't get displayed too).

observation_portal/accounts/forms.py
zemogle: The `education_user` flag is not used for any of our projects. It could be completely removed. I think this was a misunderstanding from the early Valhalla spec.

eheinrich: Oh interesting I did not know that. If we don't need it, we should probably remove that field from the Profile model entirely. I'll want to check with a couple of others before I do that, just in case. Not that doing that is necessary for this PR, just a TODO.

eheinrich (approved):

eheinrich:
I'll take a look into why the tests for this PR are not running tomorrow.

PR #206, opened 2021-07-06
Add start_exposures field for configuration status update time

jnation3406:
This adds the update exposure start time option for configuration status updates.


observation_portal/observations/serializers.py
```diff
@@ -30,6 +30,7 @@ class ConfigurationStatusSerializer(serializers.ModelSerializer):
     instrument_name = serializers.CharField(required=False)
     guide_camera_name = serializers.CharField(required=False)
     end = serializers.DateTimeField(required=False)
+    start_exposures = serializers.DateTimeField(required=False)
```
jnation3406, line 33: I named the update field this for now, but was thinking maybe it should be more descriptive of what it is, like `start_exposures_at` or `exposures_start_at`.

observation_portal/requestgroups/duration_utils.py
```diff
@@ -46,7 +46,7 @@ def get_instrument_configuration_duration(instrument_config_dict, instrument_nam
     return instrument_config_dict['exposure_count'] * duration_per_exposure
 
 
-def get_configuration_duration(configuration_dict, request_overheads):
+def get_configuration_duration(configuration_dict, request_overheads, include_front_padding=True):
```
jnation3406, line 49: Usually we want front padding when calculating configuration duration, but not when we are updating the observation end time from the exposure start time, since exposure start is after the front padding section of time.

observation_portal/requestgroups/models.py
```diff
@@ -330,6 +331,19 @@ def get_remaining_duration(self, configurations_after_priority):
             start_time,
             configurations_after_priority
         )
+        if include_current:
```
jnation3406, line 334: In this case where we want to calculate end time based on a configuration statuses exposure start time, we want to add the remaining configuration status duration to the current configurations duration of exposures (duration minus front_padding).

eheinrich (approved):

markBowman (approved):

PR #207, opened 2021-07-07
Feature/tag proposals

eheinrich:
Add tags fields to the `Proposal` and `ScienceApplication` models to store a list of strings tagging objects of those models. 

For both models, the API responses and the admin list views can be filtered by tag. Also, when a science application is ported over to a proposal, that science application's tags are added to the newly created proposal.

This branch is running here http://172.16.4.73:8000/ (corresponding frontend running here http://172.16.4.73:8080/)


observation_portal/proposals/migrations/0008_auto_20210628_0304.py
```diff
@@ -0,0 +1,24 @@
...
+        ('proposals', '0007_remove_timeallocation_instrument_type'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='proposal',
+            name='tags',
+            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(max_length=255), blank=True, default=list, help_text='List of strings tagging this proposal', size=None),
+        ),
+        migrations.AlterField(
```
eheinrich, line 19: This migration is here because it was a created from a previous update (not on this branch) that didn't have a migration along with it.

observation_portal/proposals/viewsets.py
```diff
@@ -67,6 +68,11 @@ def globallimit(self, request, pk=None):
         else:
             return Response({'errors': serializer.errors}, status=400)
 
+    @action(detail=False, methods=['get'])
+    def tags(self, request, pk=None):
```
eheinrich, line 72: This endpoint is used by the frontend to get the options for the available tags that can be filtered on.

observation_portal/proposals/admin.py
```diff
@@ -55,10 +72,11 @@ class ProposalAdmin(admin.ModelAdmin):
         'public',
         'semesters',
         'pi',
+        'tags',
         'created',
         'modified'
     )
-    list_filter = ('active', 'sca', 'public')
+    list_filter = ('active', ProposalTagListFilter, 'sca', 'public')
```
jnation3406, line 79: didn't know you could do this, cool

observation_portal/proposals/migrations/0008_auto_20210628_0304.py
jnation3406: strange, must be from me but I thought I created the migrations at the end before deploying. Oh well

observation_portal/sciapplications/filters.py
```diff
@@ -20,6 +20,7 @@ class ScienceApplicationFilter(django_filters.FilterSet):
     only_authored = django_filters.BooleanFilter(
         method='filter_only_authored'
     )
+    tag = django_filters.CharFilter(method='filter_has_tag', label='Has tag')
```
jnation3406, line 23: These might be better as a MultipleChoiceFilter? With the choices defined by the tag set for your account.

observation_portal/sciapplications/migrations/0005_auto_20210628_0304.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sciapplications', '0004_remove_timerequest_instrument'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='timerequest',
+            options={'ordering': ('semester',)},
```
jnation3406, line 16: oh I vaguely remember having to mess with the ordering so that's probably why this is here

observation_portal/sciapplications/migrations/0006_auto_20210707_0334.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('sciapplications', '0005_auto_20210628_0304'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='scienceapplication',
+            name='tags',
+            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(max_length=255), blank=True, default=list, help_text='List of strings tagging this application', size=None),
```
jnation3406, line 17: Is this migration just to fix the spelling in help text? We should probably just combine the migrations in the PR so there is just one.

jnation3406 (approved):
looks good, I can see us using this for other tags in the future

observation_portal/sciapplications/migrations/0006_auto_20210707_0334.py
eheinrich: Yeah, just a typo. Will combine.

observation_portal/sciapplications/filters.py
eheinrich: Hmm, I'm not sure if there's an easy way to get the user account of the current user in a filterset to be able to filter down the apps so I can get the list of available tags for the choices... Will look into it.

PR #209, opened 2021-07-09
allow extra params to be boolean values

jnation3406:
Before, the self_guide extra_param value was being converted to 1.0 because float(True) = 1.0. This fixes that, and also checks for string 'true', 'True', 'false', 'False' values to convert to booleans as well.


eheinrich (approved):
Looks good

PR #210, opened 2021-07-22
Fix dither expansion

jnation3406:
I found that the grid was wonky even without the rotation (try setting centering on a non-square grid to see), so I ended up rewriting the equations for the grid entirely. Both rotation and centering should now work with the grid, and the rotation is in the direction we say it should be, which is non-standard but is what Tim had in his comments so I'm sticking with it. The grid may look different then it did before, but that shouldn't matter since we can define it however we want - the grid will now go in +dec and -ra by default, but can be +dec and +ra if rotated 90 degrees and using num_rows/columns opposite. 

I also fixed the rotation in the line to be in the same direction as the grid, which is clockwise from north (+y-axis) towards east (+x-axis).


eheinrich:
Before I review, what are all the non-dither related changes in here?

jnation3406:
> Before I review, what are all the non-dither related changes in here?

Oh no! somehow my changes for updating libraries that I thought were in a different branch leaked into here. I'll try to fix this somehow...

PR #211, opened 2021-07-23
changes to fix rotation and centering in grid/line expansion

jnation3406:
Okay try two! I found that the grid was wonky even without the rotation (try setting centering on a non-square grid to see), so I ended up rewriting the equations for the grid entirely. Both rotation and centering should now work with the grid, and the rotation is in the direction we say it should be, which is non-standard but is what Tim had in his comments so I'm sticking with it. The grid may look different then it did before, but that shouldn't matter since we can define it however we want - the grid will now go in +dec and -ra by default, but can be +dec and +ra if rotated 90 degrees and using num_rows/columns opposite.

I also fixed the rotation in the line to be in the same direction as the grid, which is clockwise from north (+y-axis) towards east (+x-axis).

### DIFF
--- a/observation_portal/requestgroups/dither.py
+++ b/observation_portal/requestgroups/dither.py
@@ -54,7 +54,7 @@ def calc_line_offsets(num_points, point_spacing, orient, center):
     for i in range(0, max(num_points, 0)):
         distance = i*point_spacing + distance_offset
         # Angles measured clockwise from North ("y-axis") rather than anti-clockwise
-        # from East ("x-axis") so sin/cos flipped compared to normal
+        # from East ("x-axis")
         x_offset = (distance * -sino)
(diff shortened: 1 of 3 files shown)


jnation3406:
I added a commit to allow multiple instrument_configs in the expansion and tile it like Tim said - cycle through all configurations for each offset before moving to the next offset.

eheinrich (approved):
Looks good. The API docs might need to be updated if we wrote in there that you can't make a dither pattern with more than one instrument config.

jnation3406:
> Looks good. The API docs might need to be updated if we wrote in there that you can't make a dither pattern with more than one instrument config.

Okay I'll take a look - It probably should be updated either way to explain the behaviour when you have more than one instrument_config now.

PR #212, opened 2021-08-10
Mosaic expansion

jnation3406:
Add ability to get orientation for instrument_types with averaging, and refactor dithering expansion and add mosaicing expansion functions and serializer.


observation_portal/requestgroups/pattern_expansion.py
```diff
@@ -38,10 +31,61 @@ def expand_dither_pattern(dither_details):
...
+    offsets = expand_pattern(expansion_details)
+
+    for offset in offsets:
+        configuration_copy = deepcopy(configuration)
+        configuration_copy['target']['dec'] += (offset[1] / 3600.0)
+        cos_dec = cos(radians(configuration_copy['target']['dec']))
+        cos_dec = max(cos_dec, 10e-4)
+        configuration_copy['target']['ra'] += (offset[0] / 3600.0 / cos_dec)
+        extra_params = configuration_copy.get('extra_params', {})
+        extra_params['obs_recipe'] = 'Mosaic'
```
eheinrich, line 57: Not sure if this would make more sense at the request level extra_params. Unless we end up having requests where some configurations are part of the mosaic and some aren't

observation_portal/requestgroups/pattern_expansion.py
```diff
@@ -38,10 +31,61 @@ def expand_dither_pattern(dither_details):
...
+
+    request_dict['configurations'] = configurations
+    return request_dict
+
+
+def expand_pattern(expansion_details):
+    '''
+    Takes in just a set of pattern expansion parameters and returns a set of ra/dec offsets following
+    the pattern parameters provided
+    :param expansion_details: a valide dictionary containing a number of pattern expansion
```
eheinrich, line 69: valid

observation_portal/requestgroups/serializers.py
```diff
@@ -863,16 +862,77 @@ class DitherSerializer(serializers.Serializer):
...
+            if 'line_overlap_percent' not in validated_data:
+                validated_data['line_overlap_percent'] = validated_data['point_overlap_percent']
+            validated_data['point_spacing'] = abs(rotated_ccd_y) * pixel_scale * ((100.0 - validated_data['point_overlap_percent']) / 100.0)
+            validated_data['line_spacing'] = abs(rotated_ccd_x) * pixel_scale * ((100.0 - validated_data['line_overlap_percent']) / 100.0)
+            if flip:
+                # If the pattern orientation is closer to 90 or 270 (within 45 degrees), then flip the point/line spacing to better align with pattern orientation
+                temp = validated_data['line_spacing']
+                validated_data['line_spacing'] = validated_data['point_spacing']
+                validated_data['point_spacing'] = temp
+        elif 'point_spacing' not in validated_data:
```
eheinrich, line 925: In the case that someone sets `point_spacing` instead of the `point_overlap_percent`, should we do something similar to line 916 and also set `line_spacing` to the provided `point_spacing` if it is unset?

observation_portal/requestgroups/viewsets.py
```diff
@@ -254,6 +254,17 @@ def observations(self, request, pk=None):
             return Response([o.as_dict(no_request=True) for o in observations if o.state != 'CANCELED'])
         return Response([o.as_dict(no_request=True) for o in observations])
 
+    @action(detail=False, methods=['post'])
+    def mosaic(self, request):
+        # Check that the mosaic parameters specified are valid
+        mosaic_serializer = import_string(settings.SERIALIZERS['requestgroups']['Mosaic'])(data=request.data)
+        if not mosaic_serializer.is_valid():
+            return Response(mosaic_serializer.errors, status=400)
```
eheinrich, line 262: Can use `status=status.HTTP_400_BAD_REQUEST` here, where the import is `from rest_framework import status`

observation_portal/requestgroups/viewsets.py
```diff
@@ -254,6 +254,17 @@ def observations(self, request, pk=None):
             return Response([o.as_dict(no_request=True) for o in observations if o.state != 'CANCELED'])
         return Response([o.as_dict(no_request=True) for o in observations])
 
+    @action(detail=False, methods=['post'])
```
eheinrich, line 257: The detail-oriented part of me is slightly put off by the fact that the mosaic expansion endpoint is `/api/requests/mosaic/` but the dither expansion endpoint is `/api/configuration/dither/` (plural vs. not). What do you think about making the endpoints from the configuration viewset start with `/api/configurations/` https://github.com/observatorycontrolsystem/observation-portal/blob/main/observation_portal/urls.py#L46, like all the others?

eheinrich (approved):
Looks good. Is there also something in the LCO observation portal that only allows mosaicing for imagers?

observation_portal/requestgroups/pattern_expansion.py
jnation3406: hmm yeah im not sure, its technically possible. I wonder if we should brand obs_recipe as more of a pass-through, so it can be used for mosaic in this case but other things could be stored in there (cadence perhaps?). So I guess it would make sense in the Request for mosaics/cadences... or if it included any dither info it would need to be in the configuration.

observation_portal/requestgroups/serializers.py
jnation3406: Thats actually already in the base PatternExpansionSerializer validate method - it will set the line_spacing to point_spacing if there is point_spacing and the pattern is grid. So we only need to worry about the overlap_percent fields in the MosaicSerializer.

observation_portal/requestgroups/viewsets.py
jnation3406: okay, went ahead and switched all 400's to that within that file.

observation_portal/requestgroups/viewsets.py
jnation3406: hmm yeah I made it singular to make clear it should only take a single configuration, and we at least have one other singular endpoint with configurationstatus... But I get that the mosaic one on requests also only expects a single request. I'm fine making in plural, that'll at least be less confusing if we ever want to offer a configuration list/detail view.

observation_portal/requestgroups/pattern_expansion.py
jnation3406: oh silly me, there isn't a request level extra_params right now...

eheinrich:
> That restriction is actually in the MosaicSerializer itself because I think it would apply to anyone for the mosaic pattern expansion (pattern expansion depends on an ICRS target for mosaics since its adding to the ra/dec fields).

Agreed, but what about restricting to imager instrument types, not target types?

observation_portal/requestgroups/pattern_expansion.py
eheinrich: Oh right, for some reason I also thought we had an extra params at that level... Ok well thinking about it more, the idea of potentially having a passthrough for something like this sounds reasonable, but I guess then we wouldn't be able to set when something is both a mosaic and a cadence, it could get messy just using one field, and I'm not sure we even need to go that far.... maybe, but the use cases aren't obvious to me at this moment. Let's just stick with this and if need arises for passing though more general observing mode/ observing recipes/ whatever, then we can revisit.

PR #214, opened 2021-08-19
Add a caching decorator to locally cache most of the functions called…

jnation3406:
… many times during validate

locmem caching uses local memory, so this will increase the memory used by the obs portal pod, but I kept everything being cached to 5 minutes so we probably won't get enough unique requests in a 5 minute time to use up too much memory. And only a few functions are cached per request.

I picked these functions to cache based on what was using the most time and called the most number of times in the profiling of the /validate method. It's also sped up the /airmass method similarly, which uses a lot of the same calls. Without this caching, creating a 100 target mosaic would take ~45 seconds to validate and airmass on my machine locally, but now with the changes its more like 1.5s. The displaying of the request with 100 targets is still really taxing on my machine at last and lags using the UI at all, but it's possible to expand and submit at least within a minute. I am also minimizing all the configuration details after expansion in the UI, but that doesn't seem like it has that much of an effect.


eheinrich (approved):
Nice, looks good!

PR #215, opened 2021-08-23
Add preemption_enabled flag to cancel endpoint to allow cancelling of…

jnation3406:
… in_progress observations. I ended up leaving the logic mostly as is, except for separating the by id and by time-range, and exiting out of the cancel call early with a 400 if it will cancel an overlapping in_progress observation and doesn't have preemption_enabled flag set to True.


eheinrich (approved):
Looks good

PR #216, opened 2021-09-01
Feature/ observing mode parameters

eheinrich:
Save expansion pattern and parameters in the submitted requestgroup.

It was decided that if one instrument config has a dither, then all instrument configs within the configuration are part of a dither sequence. Similarly, if one configuration is part of a mosaic, then all configurations within that request are part of the mosaic. In addition, besides saving the mosaic and dither patterns, it was requested that the parameters used to generate the dither/ mosaic be saved in the requestgroup so that that information would be stored somewhere for informational purposes. I think this request is still up for discussion before we merge this branch in as storing the parameters in the request might not be desired since we actually don't need them in any pipelines, it would just be to record the parameters which can be done in a script or maybe the requestgroup name.

This PR adds functionality to:
- Save mosaic pattern and parameters in the request level extra_params
- Save dither pattern and parameters in the configuration level extra_params
- Validate that any dither or mosaic pattern that is set is a valid pattern
- Detect if a configuration contains a dither sequence and if so, sets that the dither pattern is a custom type.

For a mosaic, this would be stored:
```
...
"request": {
    "extra_params": {
        "mosaic_pattern": "line",
        "mosaic_pattern_params": {
            "num_points": 3,
            "orientation": 0,
            "center": false,
            "point_overlap_percent": 10,
            "line_overlap_percent": 10,
            "point_spacing": 1578.7007999999998,
            "line_spacing": 1052.4672
        }
    }
    ...
}
...
```

Technically a user can manually set the pattern to be something like `line`, but actually submit a manually created spiral pattern, which would be confusing. I don't think we can really check for this though. Let me know if you have any ideas.


observation_portal/requestgroups/pattern_expansion.py
```diff
@@ -54,11 +64,18 @@ def expand_mosaic_pattern(expansion_details):
...
 
     request_dict['configurations'] = configurations
+    # Save some information about the mosaic pattern inside the request
+    save_mosaic_params = expansion_details.copy()
+    save_mosaic_params.pop('request', None)
+    pattern = save_mosaic_params.pop('pattern', settings.MOSAIC['custom_pattern_key'])
+    if 'extra_params' not in request_dict:
+        request_dict['extra_params'] = {}
+    request_dict['extra_params']['mosaic_pattern'] = pattern
+    request_dict['extra_params']['mosaic_pattern_params'] = save_mosaic_params
```
jnation3406, line 78: I believe the mosaic pattern params will have both the `*_overlap_percent` and `*_spacing` values set since we expand the percent into a value within the serializer before it gets here. Not sure if we want to "save" both since the user only inputs one of them...

observation_portal/requestgroups/serializers.py
```diff
@@ -469,6 +469,34 @@ def validate(self, data):
                     'You may only specify a repeat_duration for REPEAT_* type configurations.'
                 ))
 
+        # Validate dither pattern
+
+        # First check that any dither pattern that is set is valid
+        dither_pattern_is_set = False
+        if 'extra_params' in data and 'dither_pattern' in data['extra_params']:
```
jnation3406, line 476: Should we also make sure that they don't set `dither_pattern` if the request doesn't have any offset values within it?

jnation3406 (approved):
Looks good. I am hoping we can convince them that we do not need to store ALL the dither or mosaic parameters. My argument is that the dither or mosaic parameters are just used in a convenience function to help a user generate a request - they are not actually related to the request itself after that so they shouldn't be stored within the request, especially in the extra_params. If anything I think I would want them stored in a separate, non-serialized field which is basically like notes on how the request is generated (and could include stuff like if it was submitted via API or the UI as well...). But really I think it just isn't necessary and is potentially harmful bloat so its best to leave it out.

observation_portal/requestgroups/serializers.py
eheinrich: Good idea, I'll add that

observation_portal/requestgroups/pattern_expansion.py
eheinrich: Removing pattern params since we decided not to store them

PR #217, opened 2021-09-02
Maintain consistent use of timezone-aware datetimes

timbeccue:
This fixes a runtime warning caused by mixing datetime objects with and without timezone awareness. Found while integrating dramatiq in the ocs_example project (while running the task `delete_old_observations`):

`dramatiq_worker_1          | /usr/local/lib/python3.7/site-packages/django/db/models/fields/__init__.py:1427: RuntimeWarning: DateTimeField Observation.end received a naive datetime (2021-08-19 20:00:00.496106) while time zone support is active.`

The `make_aware` function will use the timezone defined in `settings.TIME_ZONE`, which is currently UTC.

The task worked fine despite the warning, so this inclusion won't change program behavior. 

See https://stackoverflow.com/a/45080605/5063889

### DIFF
--- a/observation_portal/observations/tasks.py
+++ b/observation_portal/observations/tasks.py
@@ -1,7 +1,7 @@
 import dramatiq
 import logging
-from datetime import datetime, timedelta
-from django.utils.timezone import make_aware
+from datetime import timedelta
+from django.utils import timezone
 
 from observation_portal.observations.models import Observation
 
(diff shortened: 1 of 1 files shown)


eheinrich (approved):
Nice, thanks for fixing this! In other places in the code, we handle this by doing 
```

timezone.now() # Returns a timezone aware datetime of now
```
My only comment would be switch to doing it that way to keep consistent with the rest of the codebase and to skip an extra step, but not necessary.

PR #218, opened 2021-09-08
Document High Priority Endpoints

mgdaily:
So this is a big PR. I've tried to add comments to highlight the patterns I landed on and the overrides I created. Please let me know if you'd like to talk it through and I can zoom with you!

The large number of line changes are mostly due to a new configdb dump for doc generation that's a copy of the current one that's present in the repo - I wanted a separate one so we could tailor it to the docs and remove LCO-specific things, but haven't quite done that yet. Wanted to get this PR out for review so you can see the changes I'm proposing.

DRF's autodocumentation still leaves a lot to be desired, but it's decently overrideable. Thanks for your patience with this story - it's been quite challenging but I think we've finally landed on something that works decently well.

* I've added serializers where needed and we now have the ability to define separate response and request serializers for a view. This is handy when a request's body is drastically different from its response. This has actually been [introduced to the master branch of DRF](https://github.com/encode/django-rest-framework/pull/7424) but has not been included in a release.
* We can also override the request/response bodies presented in the docs - this is fragile, but in the case where a serializer cannot be written for a given endpoint ([see my StackOverflow question for an example](https://stackoverflow.com/questions/68960370/validating-nested-dictionaries-containing-dynamic-keys-using-drf-serializers)) it's quite handy.
* Added a mixin `GetSerializerMixin` for the various `APIView` classes so that we can define a serializer via `serializer_class = ...` and DRF will pick it up. `APIView` does not implement `get_serializer_class()` by default, which is what DRF schema generator uses to get the serializer for the view.
* Added an override for the name of a particular endpoint (the name that shows on the sidebar in the docs) since DRF does some silly things sometimes - any view that implements `get_endpoint_name()` can override its DRF-generated endpoint name.
* Re-wrote a lot of serializer logic to use new request and/or response serializers to validate the request and response bodies.
* Hid endpoints that aren't currently correctly documented. This [google doc](https://docs.google.com/document/d/1mBqRNe9rObVQieznDPBFOqk30nO7CY8vzKO079amQ7I/edit?usp=sharing) is up-to-date with our current state.

I have this branch building and pushing docs to https://mgdaily.github.io/observatorycontrolsystem.github.io/assets/html/observation-portal.html - that page will change as this branch is updated.

There are a number of commits here that I didn't author, but Jon did early last sprint before we started work on this in earnest. If he could comment on these that'd be great since I'm not sure what they're about - my guess is they're fallout from some library version changes that were made?


observation_portal/accounts/forms.py
```diff
@@ -27,7 +27,7 @@ class Meta:
         }
 
     def save(self, commit=True):
-        new_user_instance = super().save(commit)
+        new_user_instance = super().save(commit=True)
```
mgdaily, line 30: This change is from this commit by Jon: https://github.com/observatorycontrolsystem/observation-portal/commit/d38870f9dfac56416ea4dfe26617549338ceee2c

observation_portal/accounts/oauth2_migrations/0002_auto_20210715_0630.py
```diff
@@ -0,0 +1,74 @@
+# Generated by Django 3.2.5 on 2021-07-15 06:30
```
mgdaily, line 1: Change introduced in https://github.com/observatorycontrolsystem/observation-portal/commit/d38870f9dfac56416ea4dfe26617549338ceee2c

observation_portal/common/mixins.py
```diff
@@ -71,3 +71,43 @@ def to_internal_value(self, data):
                     except (ValueError, TypeError):
                         pass
         return data
+
+class GetSerializerMixin(object):
```
mgdaily, line 75: I created this mixin to be used by the `APIView` classes so that we could define a `serializer_class` attribute and DRF autodocs would pick it up as the serializer for that view. The base APIView does not include this functionality by default.

observation_portal/observations/models.py
```diff
@@ -1,5 +1,4 @@
 from django.db import models
-from django.contrib.postgres.fields import JSONField
```
mgdaily, line 2: Not sure about all of these changes - I am guessing they're from the dependency upgrades? I'll let Jon comment

observation_portal/common/schema.py
```diff
@@ -0,0 +1,187 @@
+from rest_framework import serializers
+from rest_framework.schemas.openapi import AutoSchema, SchemaGenerator
+from rest_framework.schemas.utils import is_list_view
+from rest_framework.viewsets import GenericViewSet
+from setuptools_scm import get_version
+from setuptools_scm.version import ScmVersion
+
+
```
mgdaily, line 8: This file contains our overrides for the base DRF doc generation functionality - I have tried to comment it as much as I can

In a nutshell:
* We can exclude actions in a viewset that we don't want documented
* We can denote a endpoint as having an empty request body - particularly useful for endpoints like "revoke token" or "accept terms"
* If the DRF autodocs are incorrectly interpreting a view as a list view, we can tell it not to via a constructor arg.
* We can also override the request/response examples with our own Python dictionaries
* If a view implements `get_request_serializer()` and/or `get_response_serializer()` the `ObservationPortalSchemaGenerator` class can and will recognize those and generate example requests and responses from them - see comment and related PR for more info.

observation_portal/proposals/serializers.py
```diff
@@ -46,6 +46,10 @@ def validate_emails(self, emails):
         return emails
 
 
+class ProposalInviteResponseSerializer(serializers.Serializer):
```
mgdaily, line 49: A common pattern I adopted when we have separate request and response serializers is to denote them as such.

observation_portal/observations/viewsets.py
```diff
@@ -156,12 +168,30 @@ def create(self, request, *args, **kwargs):
             cache.set(cache_key + f"_{site}", timezone.now(), None)
             return Response({'num_created': len(observations), 'errors': errors}, status=201)
 
+    def get_request_serializer(self, *args, **kwargs):
```
mgdaily, line 171: For viewsets with multiple actions/serializers I've adopted this pattern of defining a dict keyed by action and returning the proper serializer based on the action. It works well and is pretty clean, though I admit it kind of violates DRY, as we're almost always doing a `serializers.get()` with a sensible default - but since these are override method I didn't want to assume too much about what the implementation will look like for every view just to keep things flexible. I can probably break this stuff out into a mixin if we deem that necessary, though.

observation_portal/requestgroups/views.py
```diff
@@ -153,6 +182,7 @@ def get(self, request, instrument_type):
 
 class PressureView(APIView):
     permission_classes = (AllowAny,)
+    schema = None
```
mgdaily, line 185: `schema = None` in a view is used to hide it from being auto-documented.

setup.py
```diff
@@ -31,29 +31,30 @@
         'apscheduler>=3.7,<3.8',
         'boto3<2.0',
         'cerberus>1.0,<2.0',
-        'django>=2.2,<2.3',
-        'djangorestframework>=3.9,<3.10',
-        'django-bootstrap4<1.2',
```
mgdaily, line 36: A lot of these version changes aren't from me so I'm not super familiar

I added `setuptools-scm` and `uritemplate` as both are needed for our doc purposes.

observation_portal/accounts/management/commands/generateschema_mocked.py
```diff
@@ -0,0 +1,21 @@
+import os
+import json
+
+from rest_framework.management.commands.generateschema import Command as GenerateSchemaCommand
+from django.conf import settings
+import responses
+
+CONFIGDB_DOC_FILE = os.path.join(settings.BASE_DIR, 'observation_portal/common/test_data/doc_configdb.json')
```
mgdaily, line 8: Right now this configdb file is a direct copy of our test configdb file - it may be worth slimming it down a bit and removing some LCO-specific names. Can do that if we think it's a good idea.

observation_portal/common/test_data/doc_configdb.json
```diff
@@ -0,0 +1,2381 @@
+{
```
mgdaily, line 1: Added this = right now it's just a copy of the configdb dump we currently have in the repo as a proof of concept. Thinking it'd be good to slim down and de-LCO-ify it a bit ;)

observation_portal/requestgroups/models.py
```diff
@@ -461,7 +460,7 @@ class Configuration(models.Model):
                   'possible given its visibility within the observing window.'
     )
```
mgdaily, line 462: More changes I'm not sure about.

observation_portal/requestgroups/test/test_api.py
```diff
@@ -1196,7 +1196,7 @@ def test_cadence_invalid_period(self):
         bad_data['requests'][0]['cadence']['period'] = -666
         response = self.client.post(reverse('api:request_groups-cadence'), data=bad_data)
         self.assertEqual(response.status_code, 400)
-        self.assertEqual(response.json()['cadence']['period'], ['Ensure this value is greater than or equal to 0.02.'])
+        self.assertEqual(response.json()['requests'][0]['cadence']['period'], ['Ensure this value is greater than or equal to 0.02.'])
```
mgdaily, line 1199: I updated this test because I have updated the cadence endpoint to have an upper level serializer with a `requests` field:

```
class CadenceRequestGroupSerializer(serializers.Serializer):
    requests = import_string(settings.SERIALIZERS['requestgroups']['CadenceRequest'])(many=True)
```

This gives us accurate auto-documentation without changing our core API.

When this endpoint returns an error now, it has one extra level in the return dict. I am hoping this doesn't interfere with anything in the frontend?

observation_portal/requestgroups/serializers.py
```diff
@@ -849,6 +849,15 @@ def validate_requests(self, value):
         return value
 
 
+class CadenceRequestGroupSerializer(RequestGroupSerializer):
+    requests = import_string(settings.SERIALIZERS['requestgroups']['CadenceRequest'])(many=True)
+
+    # override the validate method from the RequestGroupSerializer and use the Cadence Request serializer to
+    # validate all of the Cadence Requests
+    def validate(self, data):
```
mgdaily, line 857: The reason I overrode the validate method here is because the RequestGroupSerializer was going and looking for windows in the requests, which don't normally exist for a CadenceRequestGroup. 

What this effectively does is pass on the validation to the CadenceRequest serializer (which specifically excludes windows)

This gives us the desired documentation results and validates properly. Prior to this, we didn't have any upper-level CadenceRequestGroup Serializer - we just looped through each request and validated it using the CadenceRequest serializer. This effectively does the same thing, but also allows us to properly document the request body.

observation_portal/common/test_data/doc_configdb.json
mgdaily: Edit: added a slimmed-down, non LCO version

observation_portal/accounts/forms.py
eheinrich: @jnation3406 I assume this is cause they changed it to be a kwarg. Should we instead do `new_user_instance = super().save(commit=commit)` to use the value of `commit` that is passed in? Or should it always just be `True`?

jnation3406: I believe it was causing issues with commit not being True. There was some issue where commit was not true, so the new user wasn't actually created so the next line of adding it into the profile wasn't working properly. I forget what the exact error/issue was at this point though...

observation_portal/urls.py
```diff
@@ -73,24 +74,23 @@
 ], 'api')
 
 schema_view = get_schema_view(
-   openapi.Info(
-      title="Observation Portal API",
-      default_version='v1',
-      description="Test description",
-      terms_of_service="https://lco.global/policies/terms/",
-      contact=openapi.Contact(email="ocs@lco.global"),
-      license=openapi.License(name="GPL 3.0 License"),
```
mgdaily, line 82: Oops, did not mean to remove license/ToS info. Will add back

observation_portal/requestgroups/views.py
```diff
@@ -164,16 +193,32 @@ def get(self, request):
...
-class ObservationPortalLastChangedView(APIView):
+class ObservationPortalLastChangedView(APIView, GetSerializerMixin):
     '''
         Returns the datetime of the last status of requests change or new requests addition
     '''
     permission_classes = (IsAdminUser,)
+    schema = ObservationPortalSchema(tags=['RequestGroups'], is_list_view=False)
+    filter_class = LastChangedFilter
+    filter_backends = (DjangoFilterBackend,)
+    serializer_class = import_string(settings.SERIALIZERS['requestgroups']['LastChanged'])
```
eheinrich, line 204: Same question about if this is still needed with `get_response_serializer` set

observation_portal/requestgroups/test/test_api.py
eheinrich: It shouldn't. Looks like the frontend doesn't actually display the specific error, just that something when wrong

observation_portal/requestgroups/viewsets.py
```diff
@@ -285,6 +325,7 @@ def get_queryset(self):
 
 class ConfigurationViewSet(viewsets.GenericViewSet):
     permission_classes = (IsAuthenticated,)
+    schema = None
```
eheinrich, line 328: I think we should document the dither endpoint

observation_portal/requestgroups/viewsets.py
```diff
@@ -183,36 +187,60 @@ def max_allowable_ipp(self, request):
 
     @action(detail=False, methods=['post'])
     def cadence(self, request):
+        request_serializer = self.get_request_serializer(data=request.data, context={'request': request})
```
eheinrich, line 190: I might be reading this wrong, but checking that this serializer is valid before expanding a cadence I think would mean that every request that gets passed in would have to contain a `cadence`? So that `else` on line 202 will never be reached

eheinrich (commented):
Nice, looking really good! So much effort, thanks for persisting. I'm not clicking approve yet just because I have a couple of questions, and you had a question in one of your comments that I wanted to come to a resolution first (the one about the 4xx/5xx responses).

observation_portal/sciapplications/filters.py
```diff
@@ -33,7 +33,7 @@ def filter_only_authored(self, queryset, name, value):
             return queryset
 
     def filter_has_tag(self, queryset, name, value):
-        return queryset.filter(tags__overlap=[value])
+        return queryset.filter(tags__overlap=value)
```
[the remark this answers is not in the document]
mgdaily: That was introduced in https://github.com/observatorycontrolsystem/observation-portal/commit/d38870f9dfac56416ea4dfe26617549338ceee2c by @jnation3406 . Googling around, I am honestly not sure

observation_portal/proposals/filters.py
```diff
@@ -28,7 +28,7 @@ class Meta:
         fields = ('active', 'semester', 'id', 'tac_rank', 'tac_priority', 'public', 'title')
 
     def filter_has_tag(self, queryset, name, value):
-        return queryset.filter(tags__overlap=[value])
+        return queryset.filter(tags__overlap=value)
```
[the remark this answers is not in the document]
mgdaily: Also introduced in https://github.com/observatorycontrolsystem/observation-portal/commit/d38870f9dfac56416ea4dfe26617549338ceee2c ...still googling

observation_portal/common/schema.py
```diff
@@ -0,0 +1,187 @@
...
+            response_schema = {
+                'type': 'array',
+                'items': item_schema,
+            }
+            paginator = self.get_paginator()
+            if paginator:
+                response_schema = paginator.get_paginated_response_schema(response_schema)
+        else:
+            response_schema = item_schema
+        status_code = '201' if method == 'POST' else '200'
```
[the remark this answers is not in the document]
mgdaily: Good catch! I can write an override

observation_portal/accounts/views.py
```diff
@@ -1,3 +1,4 @@
+from observation_portal.common.mixins import GetSerializerMixin
```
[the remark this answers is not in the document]
mgdaily: Ah, tried to do that but this one must have slipped through. Will fix

observation_portal/requestgroups/views.py
mgdaily: Nope - is technically redundant. For viewsets I used the `serializer_class` attribute as the "fallback default" in case we didn't define one for an action, but I suppose there's no sense in keeping with that an APIView

observation_portal/proposals/viewsets.py
```diff
@@ -34,54 +37,95 @@ def get_queryset(self):
...
 
     @action(detail=False, methods=['get'])
     def tags(self, request, pk=None):
         proposal_tags = get_queryset_field_values(self.get_queryset(), 'tags')
         return Response(list(proposal_tags))
 
+    def get_request_serializer(self, *args, **kwargs):
+        serializers = {'notification': import_string(settings.SERIALIZERS['proposals']['ProposalNotification']),
+                       'invite': import_string(settings.SERIALIZERS['proposals']['ProposalInvite']),
+                       'globallimit': import_string(settings.SERIALIZERS['proposals']['TimeLimit'])}
```
mgdaily, line 97: TODO: Clean up places where whitespace isn't consistent

observation_portal/requestgroups/views.py
```diff
@@ -82,24 +90,37 @@ def get(self, request):
...
 
 
-class AirmassView(APIView):
+class AirmassView(APIView, GetSerializerMixin):
     """
     Gets the airmasses for the request at available sites
     """
     permission_classes = (AllowAny,)
+    schema = ObservationPortalSchema(tags=['Requests'])
+    serializer_class = import_string(settings.SERIALIZERS['requestgroups']['Request'])
```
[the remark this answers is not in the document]
mgdaily: Nope - redundant

observation_portal/observations/models.py
jnation3406: The JSONField import removal is because its now included in the base django 3 models.

observation_portal/observations/models.py
```diff
@@ -110,7 +109,9 @@ class Observation(models.Model):
     def cancel(observations):
         now = timezone.now()
 
-        _, deleted_observations = observations.filter(start__gte=now + timedelta(hours=72)).delete()
+        observation_ids_to_delete = [observation.id for observation in observations if
+                                     observation.start >= now + timedelta(hours=72)]
+        _, deleted_observations = Observation.objects.filter(id__in=observation_ids_to_delete).delete()
```
jnation3406, line 114: I can't remember exactly why I made these changes but I think it was to fix an issue after updating libraries.

observation_portal/observations/viewsets.py
```diff
@@ -79,49 +84,56 @@ def filters(self, request):
...
+
+        response_serializer = self.get_response_serializer(data=obs_filter_options)
+        if response_serializer.is_valid():
+            return Response(response_serializer.validated_data, status=200)
+        else:
+            return Response(response_serializer.errors, status=400)
 
     @action(detail=False, methods=['post'])
     def cancel(self, request):
         """
```
jnation3406, line 96: just an fyi I updated this method a bit recently in main when you pull it in, hopefully it will merge smoothly

observation_portal/proposals/filters.py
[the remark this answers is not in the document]
jnation3406: I think this was due to a django_filters change where the value is now a list

observation_portal/requestgroups/models.py
jnation3406: just django3 changes

observation_portal/requestgroups/viewsets.py
jnation3406: Yeah that looks right now. I think we used to think this could take a RequestGroup with multiple requests, only one of which had a cadence block in it to get expanded, but I don't think that ever really made sense. It seems like we should just limit the input to a single request with a cadence block for this expansion endpoint

observation_portal/sciapplications/filters.py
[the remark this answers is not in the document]
jnation3406: Again, related to the django_filters version update changing the value to be a list

observation_portal/settings.py
```diff
@@ -310,10 +312,14 @@ def get_list_from_env(variable, default=None):
         'Schedule': os.getenv('OBSERVATIONS_SCHEDULE_SERIALIZER', 'observation_portal.observations.serializers.ScheduleSerializer'),
         'Observation': os.getenv('OBSERVATIONS_OBSERVATION_SERIALIZER', 'observation_portal.observations.serializers.ObservationSerializer'),
         'Cancel': os.getenv('OBSERVATIONS_CANCEL_SERIALIZER', 'observation_portal.observations.serializers.CancelObservationsSerializer'),
+        'CancelResponse': os.getenv('OBSERVATIONS_CANCELRESPONSE_SERIALIZER', 'observation_portal.observations.serializers.CancelObservationsResponseSerializer'),
```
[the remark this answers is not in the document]
jnation3406: agreed

observation_portal/proposals/viewsets.py
```diff
@@ -34,54 +37,95 @@ def get_queryset(self):
...
+                          'invite': 'createProposalInvite',
+                          'globallimit': 'setTimeLimit',
+                          'tags': 'getProposalTags'}
+        
+        return endpoint_names.get(self.action)
+
 
 class SemesterViewSet(viewsets.ReadOnlyModelViewSet):
     permission_classes = (AllowAny,)
+    schema = ObservationPortalSchema(tags=['Requests'])
```
jnation3406, line 121: I think this should be in the Proposals section, not Requests

jnation3406 (approved):
Looks good!

observation_portal/common/schema.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        This method is used to determine, among other things, the request and response bodies for a particular endpoint.
+        We override this to support specifying our own request and response bodies. Any view that implements get_example_response
+        and/or get_example_request can provide their own custom request and response bodies to display in the OpenAPI docs.
+        """
+        operations =  super().get_operation(path, method)
+        # If the view has implemented get_example_response, then use it to present in the documentation
+        if getattr(self.view, 'get_example_response', None) is not None:
+            example_response = self.view.get_example_response()
+            status_code = '201' if method == 'POST' else '200'
```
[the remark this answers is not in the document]
eheinrich: Yeah I think that sounds like a good idea

observation_portal/requestgroups/viewsets.py
eheinrich: That seems reasonable to me. Our API docs even say it's supposed to take only a single request https://developers.lco.global/#generate-cadence so hopefully that means that everyone who is using the endpoint is only submitting it with a single Request? Impossible to say for sure though

observation_portal/common/schema.py
```diff
@@ -0,0 +1,187 @@
...
+    def get_operation_id(self, path, method):
+        """
+        This method is used to determine the descriptive name of the endpoint displayed in the doucmentation.
+        Allow this to be overridden in the view - a view that defines get_endpoint_name can override the default
+        DRF naming scheme.
+        """
+        operation_id = super().get_operation_id(path, method)
+        if getattr(self.view, 'get_endpoint_name', None) is not None:
+            name_override = self.view.get_endpoint_name()
+            # For some viewsets, we may not have specified a name for an action - so guard against that
```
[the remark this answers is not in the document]
mgdaily: Actually I think having it return None is a perfectly valid use case. In the case of a viewset with multiple actions, some of which may need to be overridden (but others may not), we may not have defined an override for all possible actions. 

The way I have it implemented in multiple viewsets is:

```
    def get_endpoint_name(self):
        endpoint_names = {'cancel': 'cancelObservation',
                                         'filters': 'getObservationFilters'}

        return endpoint_names.get(self.action)
```

In the case of a `dict.get(action)` to get the name for the view None is a perfectly valid return value, and we should default to using what DRF provides.

observation_portal/common/schema.py
```diff
@@ -0,0 +1,187 @@
...
+            except AttributeError:
+                # If this view doesn't have a serializer, then we can't auto-document this endpoint
+                return None
+        else:
+            return view.get_request_serializer()
+
+    def get_response_serializer(self, path, method):
+        view = self.view
+
+        if not hasattr(view, 'get_response_serializer'):
```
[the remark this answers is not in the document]
mgdaily: Yeah, good point. Changed it to use `getattr()` since using try/except for control flow is a bit of an anti-pattern IMO and we should be consistent

observation_portal/requestgroups/serializers.py
```diff
@@ -898,6 +898,9 @@ class CadenceRequestGroupSerializer(RequestGroupSerializer):
     # override the validate method from the RequestGroupSerializer and use the Cadence Request serializer to
     # validate all of the Cadence Requests
     def validate(self, data):
+        if len(data['requests']) > 1:
```
eheinrich, line 901: I think we might as well check here that there is at least one request as well

observation_portal/settings.py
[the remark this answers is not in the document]
eheinrich: I should actually have been clear about this, sorry - I meant for this comment to apply for all new env vars where "RESPONSE" is not preceded by an underscore.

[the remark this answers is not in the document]
mgdaily: No worries, will fix!

observation_portal/requestgroups/serializers.py
mgdaily: So I initially did that, but in my testing I noticed that it's actually enforced at the field level because of the `many=True` kwarg passed into the serializer constructor.

So when sending a requestgroup with `"request": []` with the code as-is

we get

```
{
  "requests": [
    "You must specify at least 1 request"
  ]
}
```

I am happy to add a check like 

```
        if len(data['requests']) == 0 or len(data['requests']) > 1:
            raise ValidationError('Cadence requestgroups must contain a single request')
```

but in the case of an empty list we don't actually get to this condition

mgdaily: In any case,, I'll add a test to make sure we don't error out with an empty request list.

eheinrich: Ah, ok, that works then! yeah adding a test sounds good

eheinrich (approved):

PR #220, opened 2021-09-14
Added text at bottom of notification emails explaining how to unsubsc…

jnation3406:
…ribe from them


observation_portal/proposals/templates/proposals/requestfailurelimit.txt
```diff
@@ -11,3 +11,5 @@ Request url: {{ detail_url }}
 Thank you,
 
 {{ organization_name }}
+
+* You received this message because you opted to have notifications enabled. You can disable notifications globally here: {{ observation_portal_base_url }}/accounts/profile, or individually by proposal here: {{ observation_portal_base_url }}/proposals/{{ request.requestgroup.proposal }}
```
eheinrich, line 15: Does the asterisk do any formatting here?

eheinrich (approved):
Looks good! One thing I would add to the text is something saying that turning off notifications globally won't turn off per-proposal notifications - those will have to be turned off separately. I have a feeling that there might be some confusion with that.

observation_portal/proposals/templates/proposals/requestfailurelimit.txt
jnation3406: I don't believe so, just raw text. Do you think I should remove it?

eheinrich: No I think it's good to delineate that message in some way

PR #221, opened 2021-09-17
Fix/ Get requests for old semesters

eheinrich:
Even though we now validate against allowing users to submit observation requests with windows that span multiple semesters, we have some old requests in the system that do this, which was causing [`time_allocation_keys`](https://github.com/observatorycontrolsystem/observation-portal/blob/f14b0d1750bbb839268752921fcfb96b3d433de8/observation_portal/requestgroups/models.py#L313) to crash since `self.semester` was returning `None`, and that property accesses `self.semester.id`. This PR adds some extra code in to decide a semester when the windows of a semester span many semesters.

I figured it probably wasn't necessary to filter for observations that have been attempted since any that were scheduled are sufficient to determine which semester they were allowed to be scheduled in (and also since these are old the canceled observations will have been cleared out by now).

### DIFF
--- a/observation_portal/requestgroups/models.py
+++ b/observation_portal/requestgroups/models.py
@@ -307,7 +307,19 @@ class Request(models.Model):
 
     @property
     def semester(self):
-        return get_semester_in(self.min_window_time, self.max_window_time)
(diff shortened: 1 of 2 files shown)


jnation3406 (approved):
looks good to me

PR #222, opened 2021-10-04
Add docker compose

jnation3406:
This adds in a docker compose with the base obs portal, the task stuff, redis, the simply UI, and nginx. There is also an nginx config file that is mounted locally in the docker-compose.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,115 @@
+version: "2.4"
+
+services:
+  redis:
+    image: redis:3.2
+    command: ["redis-server", "--appendonly", "yes"]
+    restart: always
+    healthcheck:
+        test: ["CMD", "redis-cli", "ping"]
+        interval: 10s
+        timeout: 5s
+        retries: 30
+    ports:
+        -  6373:6379
+    volumes:
+        -  ./data/redis:/data
+
+  observation_portal:
+    image: observatorycontrolsystem/observation-portal
+    ports:
+        - "8000:8000"
+    environment:
+        - DB_HOST=http://127.0.0.1:5432
+        - DB_NAME=observationportal
+        - DB_USER=postgres
+        - DB_PASS=postgres
+        - SECRET_KEY=obs_portal_secret_key
+        - CONFIGDB_URL=http://127.0.0.1:7000
+        - DOWNTIMEDB_URL=http://127.0.0.1:7500
+        - ELASTICSEARCH_URL=
+        - DRAMATIQ_BROKER_HOST=redis
+        - REQUESTGROUP_DATA_DOWNLOAD_URL=http://127.0.0.1:8000/requestgroups/{requestgroup_id}/
+        - REQUEST_DETAIL_URL=http://127.0.0.1:8000/requests/{request_id}/
+        - SCIENCE_APPLICATION_DETAIL_URL=http://127.0.0.1:8000/apply/{sciapp_id}/
+        - OBSERVATION_PORTAL_BASE_URL=http://127.0.0.1:8000/
+        - ORGANIZATION_EMAIL={my_organization_email}
+        - ORGANIZATION_NAME={my_organization_name}
+        - ORGANIZATION_DDT_EMAIL={my_organization_directors_email}
+        - ORGANIZATION_SUPPORT_EMAIL={my_organization_support_email}
+        - ORGANIZATION_ADMIN_EMAIL={my_organization_web_admin_email}
+        - DRAMATIQ_BROKER_HOST=redis
+        - DRAMATIQ_BROKER_PORT=6379
+        - MAX_FAILURES_PER_REQUEST=3
+        - CORS_ALLOW_CREDENTIALS=true
+        - CORS_ORIGIN_WHITELIST=http://localhost:8080,http://127.0.0.1:8080
+        - CSRF_TRUSTED_ORIGINS=localhost:8080,127.0.0.1:8080       
+    mem_limit: "512m"
+    restart: always
+    command: >
+        sh -c "python manage.py migrate
+        && python manage.py collectstatic --no-input
+        && python manage.py runserver 0.0.0.0:8000"
+    depends_on:
+      redis:
+          condition: service_healthy
+
+
+  dramatiq_task_scheduler:
+      image: observatorycontrolsystem/observation-portal
+      environment:
+          - DB_HOST=http://127.0.0.1:5432
+          - DB_NAME=observationportal
+          - DB_USER=postgres
+          - DB_PASSWORD=postgres
+          - SECRET_KEY=obs_portal_secret_key
+          - CONFIGDB_URL=http://configdb:7000
+          - DOWNTIMEDB_URL=http://downtime:7500
+          - ELASTICSEARCH_URL=
+          - CORS_ALLOW_CREDENTIALS=true
+          - CORS_ORIGIN_WHITELIST=http://localhost:8080,http://127.0.0.1:8080
+          - CSRF_TRUSTED_ORIGINS=localhost:8080,127.0.0.1:8080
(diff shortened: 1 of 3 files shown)


observation_portal/settings.py
```diff
@@ -220,7 +220,6 @@ def get_list_from_env(variable, default=None):
 STATICFILES_DIRS = [os.path.join(BASE_DIR, 'static')]
 STATIC_URL = '/static/'
 STATIC_ROOT = '/static/'
-STATICFILES_DIRS = [os.path.join(BASE_DIR, 'static')]
```
jnation3406, line 223: Just deleting a duplicate line

nginx.conf
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    location /static/ {
+
+        alias /static/;
+
+    }
+
+
+}
```
eheinrich, line 33: No newline

docker-compose.yaml
```diff
@@ -0,0 +1,142 @@
...
+            - static_volume:/static
+        depends_on:
+            - observation_portal
+
+    observation_portal:
+        image: observatorycontrolsystem/observation-portal:3.5.4
+        expose:
+            - "8001"
+        environment:
+            # Note that 172.17.0.1 works for localhost on linux, but for mac you will
```
eheinrich, line 38: What exactly is 172.17.0.1? Why does 127.0.0.1 not work?

docker-compose.yaml
```diff
@@ -0,0 +1,142 @@
...
+            # need to use `host.docker.internal` instead to point to localhost.
+            - DB_HOST=172.17.0.1
```
eheinrich, line 40: Since we have to set the same values for these env vars in a few different places in this file, it might be better to read the values from environment variables or a .env config file https://docs.docker.com/compose/environment-variables/ I'm sure someone will make the mistake of updating the DB_USER or the image tag or something in one service but not the others, and then be confused for a bit.

docker-compose.yaml
```diff
@@ -0,0 +1,142 @@
...
+            - DRAMATIQ_BROKER_PORT=6379
+        mem_limit: "512m"
+        restart: always
+        command: >
+            sh -c "python manage.py rundramatiq --processes 2 --threads 4"
+        depends_on:
+            redis:
+                condition: service_healthy
+
+    # This provides a simple example frontend that can view and schedule Requests using
```
eheinrich, line 127: I'm on the fence about whether to include the frontend in here. Originally I was thinking we would just have the docker compose file for each repo contain the services necessary for running the app in that repo. I guess since for a prod deploy you'd run the frontend separately and the frontend would probably be different from the example frontend, given that this is a base for people to work from when starting their projects in prod, I'm leaning towards saying remove it from here and instead have a separate compose file in the frontend repo that runs just the frontend. What do you think?

nginx.conf
```diff
@@ -0,0 +1,33 @@
+upstream observation_portal {
```
eheinrich, line 1: I kinda want this file to be in a separate deploy directory, to keep it a bit more organized.

eheinrich (commented):
Just a few comments

docker-compose.yaml
jnation3406: 172.17.0.1 is the host machines ip in the default docker network that those docker-compose files use. If we used the host network setting then localhost or 127.0.0.1 might work, but otherwise those wont work since they are trying to find services running on the host machine which is not in the 'localhost' of the docker container.

docker-compose.yaml
jnation3406: yeah, that seems like a good idea

docker-compose.yaml
jnation3406: hmm yeah I guess that can work... I'm not sure how to setup the nginx (or two nginxes?) then to route /app and /o and /admin stuff to the backend but everything else to the frontend. Maybe we just only have an nginx on the backend and route any non-backend endpoints to the frontend ip, but that ip would need to be changed from default in the nginx conf if they run the frontend somewhere else...

nginx.conf
jnation3406: okay, should the docker-compose go in the deploy directory too?

docker-compose.yaml
eheinrich: Ah I see. Can you update the comment to include what 172.17.0.1 is? Might be useful if for some reason it's a different IP on someone else's machine, or they need to change network modes or something. Suggest: "When using the default docker network, `172.17.0.1` is the host machine's IP address when running on a linux machine. On Mac, it is `host.docker.internal`. When setting the environment variables for services that are running outside of this docker-compose stack but are running on the same machine, and that need to be accessed from within this stack, use the IP address of the host machine."

docker-compose.yaml
eheinrich: Right... I think it might just work since we do it in the ocs_example stack without adding an extra nginx there, but I'm not 100% sure.

nginx.conf
eheinrich: Yeah that sounds good to me

docker-compose.yaml
jnation3406: in the ocs_example stack, we have the api and frontend running from different ports, so no nginx is needed (and debug=True is on to host the staticfiles). But I'll see if I can work something out...

jnation3406: nevermind... i didn't realize the ocs_example_frontend bakes in nginx into its dockerfile, and then uses that to route to the backend endpoints with the environment variable set. It all works fine separately

jnation3406:
> Just a few comments

Made changes for all your comments, and made similar changes to the other projects docker-compose branches as applicable.

deploy/docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
+version: '3.7'
+
+# This docker-compose file brings up a full Observation Portal stack with Task Scheduler,
+# Task Worker, and Example Frontend. There is also an associated container of redis for
```
eheinrich, line 4: Update comment to remove example frontend and also update instructions for modifying the env variables referencing the .env file

deploy/docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+        ports:
+            - 8000:80
+        volumes:
+            - ./nginx.conf:/etc/nginx/conf.d/default.conf
+            - static_volume:/static
+        depends_on:
+            - observation_portal
+
+    observation_portal:
+        image: observatorycontrolsystem/observation-portal:3.5.4
```
eheinrich, line 34: I think it would be a good idea to also allow the user to configure the image repo and image tag from a single location, maybe from a separate env file or something

deploy/.env
```diff
@@ -0,0 +1,27 @@
...
+ORGANIZATION_NAME={my_organization_name}
+ORGANIZATION_DDT_EMAIL={my_organization_directors_email}
+ORGANIZATION_SUPPORT_EMAIL={my_organization_support_email}
+ORGANIZATION_ADMIN_EMAIL={my_organization_web_admin_email}
+DRAMATIQ_BROKER_HOST=redis
+DRAMATIQ_BROKER_PORT=6379
+MAX_FAILURES_PER_REQUEST=3
+CORS_ALLOW_CREDENTIALS=True
+CORS_ORIGIN_WHITELIST=http://localhost:8080,http://127.0.0.1:8080,http://172.17.0.1:8080
+CSRF_TRUSTED_ORIGINS=localhost:8080,127.0.0.1:8080,172.17.0.1:8080
```
eheinrich, line 27: No newline

eheinrich (approved):
Nice, using the env_file to configure the obs portal containers makes the compose file so much simpler!

deploy/docker-compose.yaml
eheinrich: Yeah should work

PR #225, opened 2021-12-21
Added tests and fixed up the authentication integration

jnation3406:
This adds `is_superuser` to the returned profile data in the API. It adds signals for watching User or Profile model updates, which then trigger a client view's users to be updated as well. This also occurs during a revoke_token API call. There is a management command that can be used to update all users in a client application, which will be useful to initially seed those applications user accounts.

### DIFF
--- a/observation_portal/accounts/apps.py
+++ b/observation_portal/accounts/apps.py
@@ -3,3 +3,7 @@ from django.apps import AppConfig
 
 class AccountsConfig(AppConfig):
     name = 'observation_portal.accounts'
+
+    def ready(self):
+        import observation_portal.accounts.signals.handlers  # noqa
+        super().ready()
--- a/observation_portal/accounts/serializers.py
+++ b/observation_portal/accounts/serializers.py
@@ -68,6 +68,7 @@ class UserSerializer(serializers.ModelSerializer):
 
     def get_tokens(self, obj):
         return {
+            'archive': obj.profile.archive_bearer_token,
             'api_token': obj.profile.api_token.key
         }
 
--- a/observation_portal/accounts/serializers.py
+++ b/observation_portal/accounts/serializers.py
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 from django.contrib.auth.models import User
-from django.utils.translation import ugettext as _
+from django.utils.translation import gettext as _
 from django.utils.module_loading import import_string
 from django.conf import settings
 
@@ -44,7 +44,7 @@ class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
         fields = (
-            'username', 'first_name', 'last_name', 'email', 'profile', 'is_staff', 'proposals',
+            'username', 'first_name', 'last_name', 'email', 'profile', 'is_staff', 'is_superuser', 'proposals',
             'available_instrument_types', 'tokens', 'proposal_notifications'
         )
 
@@ -68,7 +68,6 @@ class UserSerializer(serializers.ModelSerializer):
 
     def get_tokens(self, obj):
         return {
-            'archive': obj.profile.archive_bearer_token,
             'api_token': obj.profile.api_token.key
         }
 
@@ -78,8 +77,7 @@ class UserSerializer(serializers.ModelSerializer):
         return data
 
     def update(self, instance, validated_data):
-        user_update_fields = ['email', 'username', 'last_name', 'first_name']
-        instance.username = validated_data.get('username', instance.username)
+        user_update_fields = ['email', 'last_name', 'first_name']
         instance.email = validated_data.get('email', instance.email)
         instance.last_name = validated_data.get('last_name', instance.last_name)
         instance.first_name = validated_data.get('first_name', instance.first_name)
--- a/observation_portal/accounts/signals/handlers.py
+++ b/observation_portal/accounts/signals/handlers.py
@@ -1,0 +1,25 @@
+import json
+from django.dispatch import receiver
+from django.db.models.signals import post_save
+from django.contrib.auth.models import User
+from django.utils.module_loading import import_string
+from django.conf import settings
+
+from observation_portal.accounts.tasks import update_or_create_client_applications_user
+from observation_portal.accounts.models import Profile
+
+
+@receiver(post_save, sender=User)
+def cb_user_post_save(sender, instance, created, *args, **kwargs):
+    # Only update the user if it was not just created. If it was just created, we must wait until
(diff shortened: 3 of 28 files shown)


observation_portal/accounts/serializers.py
```diff
@@ -78,8 +77,7 @@ def validate_email(self, data):
         return data
 
     def update(self, instance, validated_data):
-        user_update_fields = ['email', 'username', 'last_name', 'first_name']
-        instance.username = validated_data.get('username', instance.username)
+        user_update_fields = ['email', 'last_name', 'first_name']
```
jnation3406, line 80: This is to stop allowing user's to update their username, since the username will be the 'fixed' field used when updating client user accounts. I think this is inline with other User management systems out there - there are other problematic issues with allowing users to change usernames since they are stored in the data files as a username.

observation_portal/accounts/signals/handlers.py
```diff
@@ -0,0 +1,25 @@
...
+from observation_portal.accounts.tasks import update_or_create_client_applications_user
+from observation_portal.accounts.models import Profile
+
+
+@receiver(post_save, sender=User)
+def cb_user_post_save(sender, instance, created, *args, **kwargs):
+    # Only update the user if it was not just created. If it was just created, we must wait until
+    # the Profile is created to send the update since we need profile information, which should happen
+    # immediately after user creation in our registration system.
+    if not created and not (kwargs.get('update_fields') == frozenset({'last_login'})):
```
jnation3406, line 17: updates to only 'last_login' are filtered out since they occur on admin interface login, and are not part of the profile API so they won't be propagated to the client apps users.

observation_portal/accounts/tasks.py
```diff
@@ -3,12 +3,31 @@
...
+    """
+    This call creates or updates the user account and profile information for each of the Oauth
+    Client applications specified in settings
+    """
+    for base_url in settings.OAUTH_CLIENT_APPS_BASE_URLS:
+        url = urljoin(base_url, '/authprofile/addupdateuser/')
+        logger.warning(f"Updating user details at {url}")
+        header = {'Authorization': f"Server {settings.OAUTH_SERVER_KEY}"}
+        response = requests.post(url, data=user_json, headers=header)
+        response.raise_for_status()
```
jnation3406, line 28: This will raise an error on connection issues, which should just trigger dramatiq to retry with a exponential falloff.

observation_portal/accounts/test_utils.py
```diff
@@ -1,18 +1,25 @@
...
 from django.utils import timezone
+from django.db.models import signals
+from django.db.models.signals import post_save
 
 from observation_portal.accounts.models import User, Profile
+from observation_portal.common.test_helpers import disconnect_signal
+from observation_portal.accounts.signals.handlers import cb_user_post_save, cb_profile_post_save
 
 
 def blend_user(user_params=None, profile_params=None):
```
jnation3406, line 11: This method is now used in most tests to blend the User/Profile, since it disables the new signals before creating so we don't gum up the tasks for those tests that depend on the tasks.

observation_portal/accounts/serializers.py
```diff
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 from django.contrib.auth.models import User
-from django.utils.translation import ugettext as _
+from django.utils.translation import gettext as _
```
jnation3406, line 3: This change is just to prepare for a future django 4 upgrade...

observation_portal/accounts/tasks.py
```diff
@@ -3,12 +3,31 @@
...
 
+@dramatiq.actor()
+def update_or_create_client_applications_user(user_json):
+    """
+    This call creates or updates the user account and profile information for each of the Oauth
+    Client applications specified in settings
+    """
+    for base_url in settings.OAUTH_CLIENT_APPS_BASE_URLS:
+        url = urljoin(base_url, '/authprofile/addupdateuser/')
+        logger.warning(f"Updating user details at {url}")
```
mgdaily, line 25: Could probably make this a logger.info?

observation_portal/accounts/test_utils.py
mgdaily: Nice!

observation_portal/accounts/tasks.py
mgdaily: I suppose the only thing I can think of to optimize this is to fork out the tasks so that each one goes to a separate client app - e.g. moving the `for base_url in settings.OAUTH_CLIENT_APPS_BASE_URLS:` loop into the signal handlers, and spinning up N tasks, one for each client app.

Given that it's not a hugely expensive request/operation it's not a huge deal, but it does seem wasteful to retry and update all apps when only a subset of them fails.

observation_portal/accounts/apps.py
```diff
@@ -3,3 +3,7 @@
 
 class AccountsConfig(AppConfig):
     name = 'observation_portal.accounts'
+
+    def ready(self):
+        import observation_portal.accounts.signals.handlers  # noqa
```
mgdaily, line 8: Not familiar with this, but I am guessing it's to enable signals and handlers in the accounts app?

mgdaily (approved):
Looks good! Just a few minor questions/comments.

observation_portal/accounts/tasks.py
mgdaily: But maybe not...I guess I could see a race condition where a user revokes their token multiple times when, say, the science archive is down and not accepting connections. If there's multiple tasks running trying to update the science archive user who knows which one will hit first when it comes back online. May be better to update them all at once like you've done, so at least all client apps will be consistent.

observation_portal/accounts/apps.py
jnation3406: yeah, its so the signal handler @receiver gets loaded in with the app

observation_portal/accounts/tasks.py
jnation3406: good idea

observation_portal/accounts/tasks.py
jnation3406: So I think I'll leave this as-is for now, since I'm not sure which approach will be better, and if we notice an issue with the authentication stuff getting out of sync due to this issue we can think about changing it later. On the one hand, I don't think this will normally be an issue, but say if the downtimeDB is taken down for a day or something I guess none of our user updates would propagate until it came back online. But probably in those cases we will furiously be trying to get back the service that is down anyway, so it shouldn't be an issue for long.

PR #227, opened 2022-02-14
Django 4 & Python 3.10

jashan-lco:


Dockerfile
```diff
@@ -1,14 +1,66 @@
...
 
-RUN apt-get update \
-  && apt-get install -y gfortran \
-  && pip install 'numpy>=1.16,<1.17'
+# use bash
+SHELL ["/bin/bash", "-c"]
 
-COPY . .
-RUN pip install django-ocs-observation-portal
+# install any secruity updates
```
mgdaily, line 9: security

observation_portal/common/schema.py
```diff
@@ -1,23 +1,15 @@
...
-    """
-    return str(version.tag)
 
 
 class ObservationPortalSchemaGenerator(SchemaGenerator):
     def get_schema(self, *args, **kwargs):
         schema = super().get_schema(*args, **kwargs)
         schema['info']['title'] = 'Observation Portal'
-        schema['info']['version'] = get_version(version_scheme=version_scheme, local_scheme='no-local-version')
+        schema['info']['version'] = metadata.version("django-ocs-observation-portal")
```
mgdaily, line 12: I remember you alluding to this - this is how we get the auto-generated docs to display the correct version based on the git tag. Does this still allow for that, or do we need to somehow inject that into the doc build GitHub workflow?

https://github.com/observatorycontrolsystem/observation-portal/blob/main/.github/workflows/docs-gh-pages.yaml

mgdaily (approved):
Looks good to me, just a question about detecting the version for the auto-docs

Dockerfile
```diff
@@ -1,14 +1,66 @@
...
+# collect all static assets into one place: /static
+RUN mkdir -p static && python manage.py collectstatic --noinput
+
+ENV PYTHONUNBUFFERED=1 PYTHONFAULTHANDLER=1
+
+
+# add a multi-stage build target which also has dev (test) dependencies
+# usefull for running tests in docker container
+# this won't be included in the final image
+# e.g. docker build --target dev .
```
jnation3406, line 47: Do we need to do --target prod to target the prod build, or is that default? If so, do we need to modify the github action build_and_publish_image to target prod there?

jnation3406 (approved):
Looks good to me

observation_portal/common/schema.py
jashan-lco: I'll test the doc generation locally to make sure, but I don't think we need any changes in that GH action.

The metadata version is coming from whatever is defined in `pyproject.toml`. That should be managed with `poetry version`. For example, to bump the patch version you'd do `poetry version patch`. Similar shortcuts exist for path, major, minor, prerelease, etc.

To actually create a release/tag, the workflow might have to change a little. You'd bump the version first and then commit and tag. This way the version number is checked into source code and is in that tag. E.g:
```
$ poetry version premajor # bump the version
$ git add pyproject.toml && git commit -m "version $(poetry version --short) # commit the version bump
$ git tag $(poetry version --short)
$ git push --tags
```
I can add this to the docs.

But if this is too cumbersome, I can look into other options too.

.github/workflows/build.yaml
```diff
@@ -70,15 +72,13 @@ jobs:
...
-      env:
-        TWINE_USERNAME: __token__
-        TWINE_PASSWORD: ${{ secrets.PYPI_OBS_PORTAL_API_TOKEN }}
       run: |
-        python setup.py sdist bdist_wheel
-        twine upload dist/*
-      
+        poetry build -f wheel
+        poetry build -f sdist
+        poetry publish -u "__token__" -p "${{ secrets.PYPI_OBS_PORTAL_API_TOKEN }}"
```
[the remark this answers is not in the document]
jashan-lco: Yes, these might contain special chars (that bash might not like), so generally a good idea to quote CLI arguments. Actually, I'm going to change this to single quotes to be even more cautious.

And the reason I moved these from environment variables to template, is because I believe that GitHub actions doesn't print out templated secrets in the logs. I don't know what the behavior is for environment variables.

observation_portal/common/schema.py
mgdaily: I think that's fine, it's just a new workflow we'll have to get used to. Like anything else, document it and we'll try and make sure we adhere to it.

Dockerfile
jashan-lco: Nope, the last image is always the default in a multi-stage build.

PR #228, opened 2022-03-07
Update to use OpenSearch instead of ElasticSearch

jnation3406:


observation_portal/common/telescope_states.py
```diff
@@ -36,12 +35,12 @@ class TelescopeStates(object):
 
     def __init__(self, start, end, telescopes=None, sites=None, instrument_types=None, location_dict=None, only_schedulable=True):
         try:
-            if not settings.ELASTICSEARCH_URL:
-                raise ImproperlyConfigured("ELASTICSEARCH_URL")
-            self.es = Elasticsearch([settings.ELASTICSEARCH_URL])
+            if not settings.OPENSEARCH_URL:
+                raise ImproperlyConfigured("OPENSEARCH_URL")
+            self.es = OpenSearch(settings.OPENSEARCH_URL, http_compress=True)
```
jashan-lco, line 40: If we're planning on pointing this to the tee, we should probably disable `http_compress`. And to avoid having to change this again maybe add another `OPENSEARCH_HTTP_COMPRESS` env?

jashan-lco (approved):

observation_portal/common/telescope_states.py
jnation3406: So this and the scheduler are only readers, so I think we can point them directly to the new opensearch, not the tee. But just to note, the reports project is a reader and writer but appears to work fine with the `http_compress=True` so I've left it in that project for now too.

PR #229, opened 2022-03-18
Feature/copy sciapp

mgdaily:
Adds the ability to copy sci applications. I only have a few unit tests - I couldn't really think of too many weird edge cases. Basically just wanted to check that 1) all of the relevant info is copied and the PDF path is updated to something different 2) copying fails when a call isn't open or call type doesn't match the old application


jnation3406 (approved):
Looks good to me!

PR #231, opened 2022-04-11
Feature/add nodding

jnation3406:
Adds nodding support by adding the "configuration_repeats" field to the Request model. The duration_utils code is updated to support using this value to calculate request durations properly, and the observation / config status code is updated to support expanding the /api/schedule output to contain the configurations repeated (so no changes are needed in site software), and to calculate configuration status / observation end times by inferring which repeat you are in for the configuration status being updated. 

Here is a simple example of what happens for nodding:
User submits Request with 2 configurations (a and b), configuration_repeats of 3
Scheduler schedules an observation with 2 x 3 = 6 configuration statuses in the order ababab
Site gets /api/schedule from observation portal and sees an observation with 6 full configurations, in the order ababab


observation_portal/observations/models.py
```diff
@@ -227,7 +244,6 @@ def as_dict(self):
         return import_string(settings.AS_DICT['observations']['ConfigurationStatus'])(self)
 
     class Meta:
-        unique_together = ('configuration', 'observation')
```
jnation3406, line 230: I'm not sure why we had this constraint initially - I don't see any reason that we need it, and it is impossible if we want to allow multiple configuration statuses to point to the same configuration (which is how nodding is implemented here).

mgdaily: Yeah, that makes sense.

observation_portal/requestgroups/duration_utils.py
```diff
@@ -258,13 +258,15 @@ def get_request_duration_by_instrument_type(request_dict):
...
-    # TODO: We should move front_padding to the telescope level rather than instrument_type so we don't need to
-    # assign it's value proportionally to observation time used per instrument_type
     total_duration = 0.0
-    for duration in durations_by_instrument_type.values():
-        total_duration += duration
     for instrument_type in durations_by_instrument_type.keys():
+        # First multiply the configuration durations by the configuration_repeats count
+        durations_by_instrument_type[instrument_type] *= request_dict.get('configuration_repeats', 1)
+        total_duration += durations_by_instrument_type[instrument_type]
+    for instrument_type in durations_by_instrument_type.keys():
```
mgdaily, line 266: Maybe I'm misunderstanding, but do we need two for loops here? It seems to me we could do everything for each instrument type in a single loop? Or is there some subtlety I'm missing?

mgdaily (approved):
Nice! I like this implementation - it makes sense to me. I have a few nitpicks and a question, but otherwise looks pretty good

observation_portal/observations/models.py
```diff
@@ -26,18 +28,23 @@ def observation_as_dict(instance, no_request=False):
...
-            configuration['guide_camera_name'] = config_status.guide_camera_name
-            if hasattr(config_status, 'summary'):
-                configuration['summary'] = config_status.summary.as_dict()
-            else:
-                configuration['summary'] = {}
+        expanded_configurations = []
+        configuration_status_by_config = defaultdict(list)
+        for config_status in instance.configuration_statuses.all():
+            configuration_status_by_config[config_status.configuration.id].append(config_status)
+        for i in range(instance.request.configuration_repeats):
```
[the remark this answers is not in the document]
jnation3406: sure, I'll change to repeat_index

observation_portal/requestgroups/duration_utils.py
jnation3406: Yeah, the first loop is to populate the total_duration including all instrument types. We need the final total_duration to use in the second loop to calculate the proportional amount of time per instrument_type, which we use to give the proper proportion of front_padding time to each instrument times duration. Like the TODO comment says, if we move the front_padding from an instrument_type to the telescope level in configdb, then we wouldn't need to do these loops.

observation_portal/observations/models.py
```diff
@@ -26,18 +28,23 @@ def observation_as_dict(instance, no_request=False):
...
-            config_status = configuration_status_by_config[configuration['id']]
-            configuration['configuration_status'] = config_status.id
-            configuration['state'] = config_status.state
-            configuration['instrument_name'] = config_status.instrument_name
-            configuration['guide_camera_name'] = config_status.guide_camera_name
-            if hasattr(config_status, 'summary'):
-                configuration['summary'] = config_status.summary.as_dict()
-            else:
-                configuration['summary'] = {}
+        expanded_configurations = []
```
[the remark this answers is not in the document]
jnation3406: sure, I'll do both of those

PR #240, opened 2022-06-14
fix: accept pending proposal invites on bulk user creation

jashan-lco:


observation_portal/proposals/models.py
```diff
@@ -317,11 +317,11 @@ def __str__(self):
         return 'Invitation for {} token {}'.format(self.proposal, self.email)
 
     def accept(self, user):
-        Membership.objects.create(
+        mem, _ = Membership.objects.get_or_create(
             proposal=self.proposal,
-            role=self.role,
             user=user,
         )
+        mem.role = self.role
```
jnation3406, line 324: Don't you need to do mem.save() to save setting the role? Maybe it would be better to just keep setting it within the get_or_create using defaults={'role': self.role}

jnation3406 (approved):
Mostly looks good to me

observation_portal/proposals/models.py
jashan-lco: Not sure, none of the tests complained, but perhaps we don't have tests around this.

If I understand `defaults` correctly, it'll only be used when creating a new membership. I want to update the role always so that the last call `accept` is the final role in the db.

PR #252, opened 2023-01-27
Feature/refund button

jnation3406:
This adds a variable `time_charged` to the ConfigurationStatus model to store the hours of time charged to a TimeAllocation for that configuration status - it is incremented when time is charged, which is when a Summary is saved. 

There are functions to refund a percentage of a the time used for configuration statuses, or for observations, and that both refunds the TimeAllocation and reduces the `time_charged` variable.

There are buttons added to the Observation and ConfigurationStatus detail admin pages to refund a percentage of time from those. It is book kept in the `time_charged` field to prevent refunding more time than is charged. The controls are shown below:
refund_controls


observation_portal/observations/admin.py
```diff
@@ -21,13 +21,31 @@ class ObservationAdmin(admin.ModelAdmin):
...
 class ConfigurationStatusAdmin(admin.ModelAdmin):
     list_display = ('id', 'state', 'instrument_name', 'guide_camera_name')
     list_filter = ('state', )
     raw_id_fields = ('observation', 'configuration')
     inlines = [SummaryInline]
+    change_form_template = "observations/observation_changeform.html"
+
+    def response_change(self, request, obj):
+        if "refund-time" in request.POST:
+            refund_percent = int(request.POST['refund-percent'])
```
mgdaily, line 44: Might be overkill but should we do some validation on the percentage ([0, 100])? Though I see you did it in the html form

mgdaily (approved):
Looks good, a few comments and a question!

observation_portal/observations/admin.py
jnation3406: since its in the form I don't think its possible to submit the form with a number out of that range...

observation_portal/observations/templates/observations/configurationstatus_changeform.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'admin/change_form.html' %}
+
+{% block submit_buttons_bottom %}
+    {{ block.super }}
+    <div class="submit-row">
+        <label for="refund-percent">Refund Percent:</label> 
+        <input type="number" min="0" max="100" value="100" step ="1" id="refund-percent" name="refund-percent" style="width:50px;height:16px;">
+        <input type="submit" value="Refund Time" name="refund-time">
+    </div>
+{% endblock %}
```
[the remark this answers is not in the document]
jnation3406: I realized the forms are the same so I can just remove this one, since the observation changeform can be used for both places

PR #253, opened 2023-01-27
alllow controllling error emails with environment variable

jashan-lco:


jnation3406 (approved):
I was thinking we could just unset ORGANIZATION_ADMIN_EMAIL to stop the emails, but that has other knock on effects and might still spam failed emails so this is better. Not sure about the bool(int()) casting though... If someone puts "True" in the env variable it'll crash with a ValueError on the int cast, and I can see someone thinking it would accept true/false. Some of our other projects like configdb have copied a str2bool function into the settings file (which probably originated in distutils.util.strtobool()), which accepts strings of any truthy thing or falsey thing, like 't', 'true', 'yes', '1'...

PR #259, opened 2023-02-17
Respect active flags in configdb, and update telescope states api to …

jnation3406:
…show non-schedulable instruments as well

This is for the issue mark/edward reported today. The change to respecting the configdb active flag on site/enclosure/telescope will affect ALL things using configdb data in the portal, but I think thats the right thing to do, rather than only use the instrument state as the determiner.

The other knock on effect is that the telescope states api will show telescope states for telescope with non-schedulable instruments. This might be misleading for some users since their request can't actually happen on that telescope at that moment, but the api is just showing the "telescope state", not the "instrument state" I guess. Looking at telescope states in the past was already misleading because it uses current configdb data to get the list of telescopes to show in the past, which isn't necessarily accurate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ This project adheres to semantic versioning.
 
 ### Removed
 
+## [4.4.9] - 2023-02-28
+
+### Added
+- Support for validating instrument configurations via validation schema defined in the Configuration Database's ConfigurationTypeProperties.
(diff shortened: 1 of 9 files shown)


mgdaily (approved):
Looks good, I think this is the right thing to do. Like you say, it'll force us to correctly report the active state of a network resource.

PR #260, opened 2023-02-27
Add ConfigurationTypeValidationHelper class

mgdaily:
This is meant to validate instrument configs based on the validation schema specified in the ConfigurationTypeProperties model for that particular instrument type/configuration type.


jnation3406 (approved):
This looks good, I would just integrate this validation check into the ConfigurationSerializer around line 385 just like the InstrumentType validation stuff is applied. That way it'll apply to everyone that fills in validation_schema.

PR #262, opened 2023-03-28
Update epoch of perihelion validation range to allow lower values tha…

jnation3406:
…t are valid for some distant non-sidereal objects


mgdaily (approved):
lgtm

jchate6 (approved):
Doesn't address the fact that hidden fields failing validation can be hard to diagnose, but expands the range for which validation might be broken to include a few extra decades.
Looks good.

PR #263, opened 2023-03-28
Add tags to proposal section of timeallocations action on semester serializer

mgdaily:
This is so that we can display them in the `/proposals/semesteradmin` table - requested by Nikolaus/Lisa

You can see the tags are now included in the semesteradmin table at http://observation-portal-dev.lco.gtn/proposals/semesteradmin/2022A

Will open a frontend PR shortly.


jnation3406 (approved):
Looks good to me

PR #272, opened 2023-09-07
Fix Telescope States and Telescope Availability - No OS Connection

mgdaily:
We should return a useful message, with an appropriate HTTP status code when OpenSearch cannot be reached. The status code 502 [indicates](https://www.rfc-editor.org/rfc/rfc9110.html#name-502-bad-gateway) "...that the server, while acting as a gateway or proxy, received an invalid response from an inbound server it accessed while attempting to fulfill the request."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ This project adheres to semantic versioning.
 
 ### Removed
 
+## [4.8.0] - 2023-09-07
+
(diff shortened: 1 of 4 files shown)


jnation3406 (approved):
That looks good to me! I'm guessing you might also need a patch for the tom toolkit to handle the 502 still

PR #275, opened 2023-11-08
Fix registration URL

mgdaily:
I really don't understand this one, but apparently urllb stopped playing nicely with us when we had double-slashes. Perhaps this never worked?

Users were getting URLs of the form https://observe.lco.global/?email=wesleyotieno2001@gmail.com instead of https://observe.lco.global/accounts/register/?email=wesleyotieno2001@gmail.com

In any case, I've updated a test too. I'm a little busy today but can we coordinate a deploy sometime?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -51,6 +51,10 @@ jobs:
         DB_PORT: 5432
     - name: Generate and send coveralls report
       uses: coverallsapp/github-action@v2
+      with:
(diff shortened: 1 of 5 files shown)


jnation3406 (approved):
Looks good. Clearly our understanding of how urljoin works is wrong but this seems to fix it in this case. I think the query param with the slash before is what confuses it - not that it is a urljoin bug, more like we are misusing it...

PR #292, opened 2024-04-09
feat: add skaffold & kustomize resources

jashan-lco:


jashan-lco:
@jnation3406 See if you can follow `DEVELOPMENT.md` to get up and running with Skaffold.

DEVELOPMENT.md
```diff
@@ -0,0 +1,27 @@
...
+Activate development environment:
+
+```sh
+nix develop --impure
+```
+
+Create an ephmeral Kubernetes cluster for development:
+
+```sh
+ctlplt apply -f ./local-registry.yaml -f ./local-cluster.yaml
```
jnation3406, line 12: typo ctlptl

DEVELOPMENT.md
```diff
@@ -0,0 +1,27 @@
...
+skaffold -m deps run
+```
+
+Start application development loop:
+
+```sh
+skaffold -m app dev --port-forward
+```
+
+You can customize environment variables in `./k8s/envs/local/{settings,secrets}.env`.
```
jnation3406, line 27: I would include an explanation on where the service will be running after this command. i.e. `http://0.0.0.0:8080`, and that you can access the /admin interface with the default admin:admin user/pass.

jnation3406: Also some explanation on the commands to take things down after development? I got everything running with these commands and you can ctrl+c out of the final skaffold command, but that leaves two docker containers running on my machine so I'm assuming the local registry/cluster is still running via docker. So the command to take that down would be good for me and future developers.

jashan-lco: Sounds good. You can delete the cluster with `ctlptl delete -f ./local-cluster.yaml`. The registry should be left running because it's intended to be shared across projects/cluster. It caches the container images allowing us to skip big expensive container builds (e.g. banzai).

PR #296, opened 2024-05-27
Add a new REAL_TIME observation type and entrypoint for submitting re…

jnation3406:
…al time blocks

The entrypoint is `/api/realtime/` and you can post observation blocks with no request, which creates an associated observation and requestgroup of observation_type REAL_TIME, but no configurations are in it (and therefor no configuration statuses are created either). You would then view them in the /api/observations or /api/schedule GET requests with filters. 

I've added basic serialization rules for the realtime blocks, to reject those that overlap existing downtime or aren't in the telescopes nighttime.

I've tested this locally so far and can create real time blocks, and viewing them in the UI seems to work ok. My next step will be to deploy it to the dev cluster and check it with running the scheduler.


jnation3406:
Just remembered I still have to add a mechanism to set/delete downtime blocks here like we discussed

mgdaily (approved):
Nice! I really don't have much feedback - this all makes a lot of sense. I have a couple of small, pedantic comments.

I think now that we're doing this with proposals we probably need to think of the timeallocation policy for education users.

So we'll start each user with one long-lived proposal that, for instance, gets an automatic time allocation for real time observing, say, once they get far enough along in the course? And we'll probably want to somehow have a hook to roll those over into a new TimeAllocation for the next semester so that semester boundaries don't really matter? 

And instead of filtering DataLab by proposal, we probably now want to filter by RequestGroups that belong to the proposals a user is on? Since that will include both real-time and scheduled observations AND still allows us to support collaboration between users using the existing proposal system.

observation_portal/observations/viewsets.py
```diff
@@ -49,6 +56,82 @@ def observations_queryset(request):
     ).distinct()
 
 
+class RealTimeViewSet(CreateListModelMixin, viewsets.ModelViewSet):
```
mgdaily, line 59: Oh one thing I think I forgot is that we need a way for the UI to hit an endpoint to get the set of available session slots on the network +/- 2 hours (configurable) from the time they set. So we'd need to have configuration for 

1) datetime intervals in the night (first and last 2 hours on each site?) that we allow RTI observations in 
2) a set of instrument telescope classes we allow these types of observations on

Maybe that's outside the scope of this? Or would it go better in the LCO obs portal project? We could definitely get something working now without all of this, but I think an availability endpoint that we can add more complexity to would be nice.

jnation3406: Yeah I didn't do that yet since we haven't really defined what we want from it... I kind of want to flesh out what limits we want to place on it first, and then that might dictate where it should go (i.e. in the base if its configurable enough to be general, or in the lco project if its really specific stuff).

jnation3406:
> Nice! I really don't have much feedback - this all makes a lot of sense. I have a couple of small, pedantic comments.
> 
> I think now that we're doing this with proposals we probably need to think of the timeallocation policy for education users.
> 
> So we'll start each user with one long-lived proposal that, for instance, gets an automatic time allocation for real time observing, say, once they get far enough along in the course? And we'll probably want to somehow have a hook to roll those over into a new TimeAllocation for the next semester so that semester boundaries don't really matter?
> 
> And instead of filtering DataLab by proposal, we probably now want to filter by RequestGroups that belong to the proposals a user is on? Since that will include both real-time and scheduled observations AND still allows us to support collaboration between users using the existing proposal system.

Yeah I was thinking each LCO education user would have a LCO Education proposal made for them on account creation that they use for all LCO education stuff. Then different activities could grant them normal or realtime time in that proposal. Yes we would have to roll it over each semester - it might make sense to add in some option for time allocations to apply to all semesters instead. Then in Datalab UI it would make the most sense to filter by requestgroup name as well, but we will have to do some munging of obs portal and archive data since the archive doesn't have request_group ids or names as fields you can filter by (they aren't in the model)

PR #302, opened 2024-07-12
Fix zero division error in duration utils

jnation3406:
Added a "fli" instrument to the test config and wrote a test that generated the zerodivision exception, then modified the code to not through the exception. In this case, there was an okay formed request so I'm not sure there was really an "error" here per-se, just a problem with zero length requests.


mgdaily (approved):
Nice!

PR #305, opened 2024-07-24
Added realtime availability endpoint

jnation3406:
Basic availability endpoint that takes night time and downtime to generate availability intervals for all telescopes that a users proposals have time on.


mgdaily (approved):
Nice! Looks good

PR #306, opened 2024-08-16
Fix/more realtime restrictions

jnation3406:
This should be pretty straightforward, but basically just blocked 3 conditions for the two different cases of trying to book a session (in serializers.py) and trying to get availability intervals (realtime.py). Cases are:

1. Overlapping in time but not resource with another realtime session for that user
2. Overlapping in time and resource with an in progress observation
3. Overlapping in time and resource with a future scheduled pending TC, RR, or Direct observation

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -43,7 +43,7 @@ jobs:
         python -m pip install 'poetry'
         poetry install
       env:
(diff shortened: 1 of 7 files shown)


mgdaily (approved):
Nice! Looks great.

PR #308, opened 2024-09-28
Added new cerberus validator, changed instrument_type and configurati…

jnation3406:
…on_type validation_schemas to be applied at the Configuration level instead of Instrument Config, and changed how those validation_schemas errors are reported so they will mimic a DRF validation error on the bad field. A side benefit to this is that now things in the instrument_type validation_schema, like defocus,  will be validated on the frontend like all other fields.

[demo](https://github.com/user-attachments/assets/32720293-3ad1-4167-bbbe-24b7fdeeef15)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ This project adheres to semantic versioning.
 
 ### Removed
 
+## [4.11.0] - 2024-10-11
+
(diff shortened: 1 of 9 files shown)


mgdaily (approved):
Nice! A lot fewer changes than I realized

PR #313, opened 2024-12-17
feat: add more info to proposal review emails

jashan-lco:


jnation3406 (commented):
Looks good, I see you moved the `send_review_accepted_or_rejected_email_to_submitter` method to the admin interface since its only used there now, but I think it could make sense to stay in the model as a method too since it is business logic on that model...

jashan-lco:
> Looks good, I see you moved the `send_review_accepted_or_rejected_email_to_submitter` method to the admin interface since its only used there now, but I think it could make sense to stay in the model as a method too since it is business logic on that model...

Good point

PR #321, opened 2025-01-30
Return an error if downtime db connection fails while creating or del…

jnation3406:
…eting realtime session. Also create a configuration status for realtime observations

There's two things being fixed in this PR.
1. If the downtime DB connection fails when trying to create or delete a realtime session, it used to still create/delete the corresponding observation in the obs portal. Now it should kick back an error to the caller saying to try again, since the downtime db connection stuff is usually transitory.
2. The realtime observations didn't have configuration statuses created, since they have no configurations. But configuration statuses are necessary for the site software to report back status on the observation. The migration is just to allow null for configuration on a configuration status. I didn't want to create a useless configuration + all the stuff it needs (constraints, target, etc.) for a realtime observation since its not necessary.


observation_portal/observations/viewsets.py
```diff
@@ -82,10 +82,6 @@ def perform_create(self, serializer):
             using its observation id as the downtime reason
         """
         observation = serializer.save(submitter=self.request.user, submitter_id=self.request.user.id)
-        # Debit the realtime time allocation hours
```
jashan-lco, line 85: Would wrapping this whole thing in a database transaction help make sure there are no edge-cases with the clean-up?

observation_portal/observations/models.py
```diff
@@ -35,6 +35,9 @@ def observation_as_dict(instance, no_request=False):
 def get_expanded_configurations(observation, configurations):
     ''' Gets set of expanded configurations with configuration details filled in for a given observation
     '''
+    # If this is a REAL_TIME observation, just return its configuration_status since it doesn't have a configuration
+    if observation.request.request_group.observation_type == RequestGroup.REAL_TIME:
+        return [{'configuration_status': observation.configuration_statuses.first().id}]
```
jashan-lco, line 40: Is there something that ensures there's at least 1 configuration_statuses for a real time observation type? Otherwise, this might return an exception about accessing `.id` on `None`

observation_portal/observations/serializers.py
```diff
@@ -456,6 +457,7 @@ def create(self, validated_data):
             request_group = RequestGroup.objects.create(**validated_data)
             request = Request.objects.create(request_group=request_group)
             observation = Observation.objects.create(request=request, **obs_fields)
+            ConfigurationStatus.objects.create(observation=observation, instrument_name='')
```
jashan-lco, line 460: This might be the answer to my question above about ensuring a configuration status, right?

jashan-lco (approved):

observation_portal/observations/models.py
jnation3406: That's a good point - there will be from this PR (it is always created in the serializer now), but past REAL_TIME observations don't have them yet. I'll protect against that.

observation_portal/observations/viewsets.py
jnation3406: I think that would work, but I'm worried about having an external API call within a transaction since that seems like bad practice since the external api call could hang or talk a while and block other db writes...

PR #325, opened 2025-02-21
send single email to proposal review panelists

jashan-lco:


jnation3406 (approved):
The code changes look good to me - I didn't really look at the yaml since you know that best anyway

PR #326, opened 2025-02-26
Changed behavior of checking visibliity windows to ensure that only f…

jnation3406:
…uture windows are used when determining if the observation is visible. This check is only made in the serializer of the request, because we only care about it then to decide whether to reject or accept a request as visible. Other times we call that function, like when calculating the visibility windows to get the requests telescope states or airmass values, we want to include past dates that it was visible to show them all on the plot.

This came about because someone accidentally submitted a request with a window starting several days in the past and running into the future. It had visibility chunks in the past and in the future, but only the chunks in the past were large enough to fit the requests duration, so the request could actually never be scheduled once it was submitted. This should fix that bug by rejecting the request at submission.

### DIFF
--- a/observation_portal/accounts/serializers.py
+++ b/observation_portal/accounts/serializers.py
@@ -73,7 +73,8 @@ class UserSerializer(serializers.ModelSerializer):
                 proposal_details['time_used'] = obj.profile.time_used_in_proposal(proposal)
                 proposal_details['time_allocations'] = []
                 for time_allocation in proposal.timeallocation_set.filter(semester=proposal.current_semester):
-                    proposal_details['time_allocations'].append(time_allocation.as_dict(exclude=['semester', 'proposal', 'id']))
+                    proposal_details['time_allocations'].append(time_allocation.as_dict(
+                        exclude=['semester', 'proposal', 'id', 'instrument_type']))
             proposals.append(proposal_details)
         return proposals
 
--- a/observation_portal/observations/models.py
+++ b/observation_portal/observations/models.py
@@ -38,7 +38,11 @@ def get_expanded_configurations(observation, configurations):
     # If this is a REAL_TIME observation, just return its configuration_status since it doesn't have a configuration
     if observation.request.request_group.observation_type == RequestGroup.REAL_TIME:
         if observation.configuration_statuses.all().count() > 0:
-            return [{'configuration_status': observation.configuration_statuses.first().id}]
+            config_status = observation.configuration_statuses.first()
+            return [{
+                'configuration_status': config_status.id,
+                'state': config_status.state
+            }]
         else:
             return []
     expanded_configurations = []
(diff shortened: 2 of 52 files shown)


observation_portal/accounts/serializers.py
```diff
@@ -73,7 +73,8 @@ def get_proposals(self, obj):
                 proposal_details['time_used'] = obj.profile.time_used_in_proposal(proposal)
                 proposal_details['time_allocations'] = []
                 for time_allocation in proposal.timeallocation_set.filter(semester=proposal.current_semester):
-                    proposal_details['time_allocations'].append(time_allocation.as_dict(exclude=['semester', 'proposal', 'id']))
+                    proposal_details['time_allocations'].append(time_allocation.as_dict(
+                        exclude=['semester', 'proposal', 'id', 'instrument_type']))
```
jnation3406, line 77: This is just a little fix to a previous change I made for getting proposal details from the profile endpoint - unrelated to the visibility bug.

observation_portal/observations/models.py
```diff
@@ -38,7 +38,11 @@ def get_expanded_configurations(observation, configurations):
     # If this is a REAL_TIME observation, just return its configuration_status since it doesn't have a configuration
     if observation.request.request_group.observation_type == RequestGroup.REAL_TIME:
         if observation.configuration_statuses.all().count() > 0:
-            return [{'configuration_status': observation.configuration_statuses.first().id}]
+            config_status = observation.configuration_statuses.first()
+            return [{
+                'configuration_status': config_status.id,
+                'state': config_status.state
```
jnation3406, line 44: This is another unrelated change that adds in the 'state' for the 'configurations' section of REAL_TIME submissions.

observation_portal/requestgroups/serializers.py
```diff
@@ -739,8 +739,8 @@ def validate(self, data):
             if largest_interval.total_seconds() <= 0:
                 raise serializers.ValidationError(
                     _(
-                        'According to the constraints of the request, the target is never visible within the time '
-                        'window. Check that the target is in the nighttime sky. Consider modifying the time '
+                        'According to the constraints of the request, the target is never visible within the future '
```
jchate6, line 742: This just feels like awkward wording.
"The target will not be visible within the time window." Implies that we are only looking at the future without sounding like it was written by a bad sci-fi time cop.

observation_portal/requestgroups/serializers.py
```diff
@@ -749,7 +749,8 @@ def validate(self, data):
                 raise serializers.ValidationError(
                     (
                         'According to the constraints of the request, the target is visible for a maximum of {0:.2f} '
-                        'hours within the time window. This is less than the duration of your request {1:.2f} hours. '
+                        'hours within the future time window. '
```
jchate6, line 752: "the target will only be visible for a maximum of {0:.2f} hours within the time window"

jchate6 (changes requested):
This looks good. One suggestion on phrasing, that will also affect the tests.

jchate6 (approved):

PR #328, opened 2025-03-07
Feature/generic realtime visibility

jnation3406:
This adds another overridable method, which will be used to set up some LCO specific rules for blocking visibility windows for realtime observing.


observation_portal/settings.py
```diff
@@ -432,9 +432,11 @@ def get_dramqtiq_broker_url():
     }
 }
 
-# Overrides for Duration calculation functions
-DURATION = {
```
jnation3406, line 436: I'm putting this other overridable method into the `OVERRIDES` section since it is more generic sounding and we now have two... This will require renaming the override in our lco obs portal project as well.

observation_portal/observations/realtime.py
```diff
@@ -120,8 +121,12 @@ def get_realtime_availability(user: User, telescope_filter: str = ''):
     for resource in telescopes_availability.keys():
         telescope, enclosure, site = resource.split('.')
         intervals = filtered_dark_intervalset_for_telescope(start, end, site, enclosure, telescope)
-        # Now also filter out running obs, future high priority (TC, RR, Direct) obs, and obs overlapping in time by the user
         intervals_to_block = []
+        # Filter out according to our custom per telescope rules
+        intervals_to_block.append(import_string(settings.OVERRIDES['realtime_intervals_to_block_for_telescope'])(
```
jnation3406, line 126: This method is called to get all the possible realtime windows a user could book

observation_portal/common/rise_set_utils.py
```diff
@@ -340,7 +342,24 @@ def get_site_rise_set_intervals(start, end, site_code):
...
+        site: The site code
+        enclosure: The enclosure code
+        telescope: The telescope code
+    Returns:
+        Intervals object of time intervals to block realtime sessions.
+    """
+    return Intervals()
+
+
+def is_realtime_interval_available_for_telescope(start: datetime, end: datetime, site: str, enclosure: str, telescope: str) -> bool:
```
jnation3406, line 362: This method is called when validating a user booking a specific realtime session to make sure its a valid time.

capetillo (approved):
Nicely done, I believe this looks good!

PR #330, opened 2025-04-10
expose more proposal review fields & add user reviews endpoint

jashan-lco:


observation_portal/sciapplications/models.py
```diff
@@ -494,6 +499,8 @@ class ScienceApplicationUserReview(models.Model):
 
     grade = models.DecimalField(blank=True, null=True, default=None, max_digits=4, decimal_places=2)
 
+    created_at = models.DateTimeField(auto_now_add=True)
```
jnation3406, line 502: In other models in this project we use a field name of `created` for this rather than `created_at`. And you might want to include `modified` too as an `auto_now=True` if that is useful?

observation_portal/sciapplications/serializers.py
```diff
@@ -46,14 +46,31 @@ class ScienceApplicationReviewSerializer(serializers.ModelSerializer):
...
 
     class Meta:
         model = ScienceApplicationReview
         fields = [
           "id", "science_category", "technical_review", "status", "mean_grade", "summary",
           "title", "semester", "abstract", "pdf_url", "completed", "can_summarize", "is_primary_reviewer", "is_secondary_reviewer", "user_reviews", "my_review",
+          "long_term", "total_requested_time_0m4", "total_requested_time_1m0", "total_requested_time_2m0",
         ]
 
+    def get_total_requested_time_0m4(self, obj):
```
jnation3406, line 62: This is all specific to aperture sizes LCO uses... Since the obs portal is open source for other observatories as well I would prefer to keep code in the base obs portal generic for these types of things. Maybe instead you could make a serializer field `total_requested_time_by_aperture` which returns a dictionary keyed on aperture size with the value of the total time requested at that telescope class size?

observation_portal/sciapplications/serializers.py
```diff
@@ -107,14 +124,34 @@ class ScienceApplicationUserReviewSerializer(serializers.ModelSerializer):
     semester = serializers.SerializerMethodField()
     abstract = serializers.SerializerMethodField()
     pdf_url = serializers.SerializerMethodField()
+    application_id = serializers.SerializerMethodField()
+    science_category = serializers.SerializerMethodField()
+    mean_grade = serializers.SerializerMethodField()
```
jnation3406, line 129: These don't all need to be `SerializerMethodFields` - you can use [SlugRelatedField](https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield) instead for simple foreign key field lookups that are one level deep. It would look like this:
```
mean_grade = serializers.SlugRelatedField(read_only=True, slug_field="mean_grade", source="science_application_review")

jnation3406 (changes requested):
A few changes to keep the obs portal code more generic and consistent

jnation3406 (approved):
looks good to me

PR #336, opened 2025-07-09
Add never_cache decorator to realtime availability endpoint to preven…

jnation3406:
…t downstream browser caching since we already have server caching

[](https://forum.djangoproject.com/t/server-caching-without-browser-caching-http-headers/5284/3)


capetillo (approved):
Thank you!!

PR #337, opened 2025-07-22
Allow staff users to circumvent realtime availability per telescope r…

jnation3406:
…estrictions

This ignores the special per-telescope blocking of realtime time windows if the user is staff.


markBowman (approved):
lgtm

PR #339, opened 2025-09-18
Plumbed error strings through for full configuration validation schema

jnation3406:
The main change here allows the error to be converted from cerberus validation error format to django rest framework error format, so it shows up on the proper field erroring on the UI. 

This will allow [this issue](https://lcoglobal.redmineup.com/issues/2704) to be resolved by adding some code to the cerberus validation_schema for the nres instrument type in configdb. It is currently in the dev configdb and looks to be working will with this change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim as base
+FROM python:3.10-slim AS base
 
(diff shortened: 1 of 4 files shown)


jchate6 (approved):
Looks good.

PR #354, opened 2026-03-19
add admin action to export key data as csv

jashan-lco:


jnation3406 (approved):
Looks good. A few suggestions and things to look into but seems like it should work well.

jashan-lco:
Daniel is set on having columns per semester. That's what I've done. It could be more efficient, but since he's going to be selecting ~20 rows max I don't think it's worth it. Also, maybe Django already does some caching magic. Not sure.

jnation3406 (approved):
So this looks okay, but its still a bit confusing to me how this is going to be used. You said Daniel wants all the time allocations / requests broken out by semester, but then how will it work as a copy/paste into a spreadsheet for him? Based on his scienceapplication or proposals in the query, there will be a varying number of columns for time requests / allocations? Even looking at the same set of proposals over time, as they get more time allocations if they are a long running proposal, the output he copies will change. I guess I'm just trying to understand how this is useful to him with his spreadsheet.

Right now this applies to ScienceApplications, but the story says for Proposals too, but I don't see that here. Proposals have `realtime` time type as well as the other 3, which has its own allocation and usage per TimeAllocation. Is the proposal part for a separate PR?

jashan-lco:
> So this looks okay, but its still a bit confusing to me how this is going to be used. You said Daniel wants all the time allocations / requests broken out by semester, but then how will it work as a copy/paste into a spreadsheet for him? Based on his scienceapplication or proposals in the query, there will be a varying number of columns for time requests / allocations? Even looking at the same set of proposals over time, as they get more time allocations if they are a long running proposal, the output he copies will change. I guess I'm just trying to understand how this is useful to him with his spreadsheet.

Great questions. I've brought up the fact that the output will change, but that doesn't seem to bother him. I guess he'll just change the column reference in the main spreadsheet.

> Right now this applies to ScienceApplications, but the story says for Proposals too, but I don't see that here. Proposals have `realtime` time type as well as the other 3, which has its own allocation and usage per TimeAllocation. Is the proposal part for a separate PR?

Yes, the current implementation for ScienceApplications will probably be iterated on and then once he can actually use it I'll figure out how to do it for Proposals too. He said this regarding the use-case: 

> The use-case for the proposal output is a mess. It is primarily to track what LCO is doing in the current semester. Largely it would be ideal if there was a dashboard in the OCS that displayed a summary of the active proposals. The spreadsheet is filling a hole in the OCS report generation. There are a small number of math operations that are performed in the Proposal spreadsheet associated with changing the number of hours dynamically during the semester. A spreadsheet is probably the right place for random bits of math.

Which is basically this, https://observe.lco.global/proposals/semesteradmin/2030A, to me, so maybe we can just add a export to CSV button there.

PR #359, opened 2026-05-19
add action to export science application data to the Proposals page

jashan-lco:


observation_portal/common/admin.py
```diff
@@ -0,0 +1,67 @@
...
+      ("Title", lambda o: o.title),
+      ("PI Name", lambda o: " ".join([o.pi_first_name, o.pi_last_name])),
+      ("PI Institution", lambda o: o.pi_institution),
+      ("PI Email", lambda o: o.pi),
+      ("Tags", lambda o: "|".join(o.tags))
+    ]
+
+    timerequest_semesters_start = {}
+    for o in sciapps:
+        for tr in o.timerequest_set.all():
```
jnation3406: I think you can get them sorted by start with something like `o.timerequest_set.all().order_by('start')` to simplify this

observation_portal/common/admin.py
```diff
@@ -0,0 +1,67 @@
+from django.conf import settings
+
+from ..sciapplications.models import ScienceApplication
```
jnation3406: I prefer to stick with absolute imports like we use elsewhere in the project.

observation_portal/proposals/admin.py
```diff
@@ -4,6 +4,7 @@
 from django.utils import timezone
 from django.http import HttpResponseRedirect
 
+from ..common.admin import export_sciapps_key_data_csv
```
jnation3406: absolute import

observation_portal/proposals/admin.py
```diff
@@ -114,6 +115,18 @@ def get_actions(self, request):
             del actions['delete_selected']
         return actions
 
+    @admin.action(description="Export science application data as CSV")
+    def export_app_key_data_csv(self, request, queryset):
```
jnation3406: Are you sure they want to export the science application data and not the proposal data here? The proposal can differ from the sci application for various reasons after porting, like moving around time between different classes and whatnot, or changing priority or other things. Also, some proposals are created directly circumventing the science application process and this wouldn't work for those either.

jnation3406 (commented):
Just a few small comments, except for the proposal export I think they might want the proposal details instead of the linked science application details again (unless you have heard otherwise), because the proposal values can be changed after porting the science application.

observation_portal/proposals/admin.py
jashan-lco: I'm not sure about anything about this. I was going off the columns he requested, but I guess it makes sense to export those from the Proposal where it makes sense because they could be changed.

jashan-lco: Ok, I left this as be, but I also added an action to export these columns from the Proposal.

observation_portal/proposals/admin.py
```diff
@@ -114,6 +116,86 @@ def get_actions(self, request):
...
+            return ret
+
+        def get_tc_time(o, inst_type, semester):
+            ret = 0
+            for ta in timeallocation_by_inst_type(o, inst_type, semester):
+                ret += ta.tc_allocation
+            return ret
+
+        for semester in timeallocation_semesters:
+            for inst_type in settings.SCI_APPS_ADMIN_EXPORT_CSV_INSTRUMENT_TYPES:
```
jnation3406: Do we need a new setting for proposals vs sciapps?

observation_portal/sciapplications/admin.py
```diff
@@ -9,6 +9,7 @@
 from django.forms.models import ModelForm
 from django.shortcuts import render
 
+from observation_portal.common.admin import export_sciapps_key_data_csv
 from .models import (
```
jnation3406: Absolute import

jnation3406 (approved):
Looks good to me

observation_portal/proposals/admin.py
jashan-lco: Oh, maybe rename it to be more generic so it applies to both? I don't see a reason why you would want one to be different from the other.

observation_portal/sciapplications/admin.py
jashan-lco: Force of habit

PR #362, opened 2026-06-25
Fix error in user authentication and block porting sci applications w…

jnation3406:
…ith zero tac priority

The user api login error message is because I noticed some 500 error emails where someone was hitting it with no username/password, which crashed things. This should return a 400 error if they leave out those fields instead of crash.

The sciapp part is a simple change to display an error in the admin interface if they try to run the port action without setting tac priority to a non-zero value first.


Fingel (approved):
Looks reasonable.

jashan-lco (approved):
Looks good

PR #363, opened 2026-07-03
Adds an import CSV utility for the proposal admin

Fingel:
https://github.com/user-attachments/assets/f48bab17-c6d5-42fd-a524-fda6881d0660

Introduces django-object-actions as a new dependency.

Creates an action on the proposal list admin to import from CSV. This button runs a view which contains an upload form for the csv, as well as other fields as required. Proposals will be created and then the user is redirected back to the admin list page.

This is to save a lot of copy and paste every new semester.

The business logic is not complete:

What to do if we can't find the PI in the database? Do not create a membership, or attempt to create the user?

I'm not familiar with what a SCA really is. Can we hardcode the proposals to link to an existing one? Or does this need to be dynamic?

Any other missing information that can be inferred or supplied by the form?


pyproject.toml
```diff
@@ -45,6 +45,7 @@ django-filter = "21.1"
 django-oauth-toolkit = "^2"
 django-redis-cache = "^3"
 django-registration-redux = "^2.9"
+django-object-actions = "^5.1.2"
```
markBowman, line 48: poetry.lock needs updating

observation_portal/proposals/admin.py
```diff
@@ -237,7 +249,80 @@ def rollover_selected(self, request, queryset):
...
+            )[0],
+        )
+
+        # Attempt to add PI membership
+        email = row["PI_email"]
+        try:
+            user = User.objects.get(email=email)
+            Membership.objects.create(user=user, proposal=proposal, role="PI")
+        except User.DoesNotExist:
+            # could not find PI. Do what here? Create or no membership?
```
markBowman: We should  probably still create the proposal. SciOps would have to manually tie the proposal to the correct PI account.

markBowman: The question is whether we have a mechanism to alert the person that ran the csv import that there was a problem.

observation_portal/proposals/admin.py
```diff
@@ -237,7 +249,80 @@ def rollover_selected(self, request, queryset):
...
                 proposals.append(obj)
         return render(request, 'admin/rollover_selected.html', context={'proposals':proposals, 'rejects':rejects, 'updated':updated, 'nextsemester':nextsemester, 'currentsemester':currentsemester})
-                
+
+    @action(label="Import Proposals from CSV")
+    def import_proposals_csv(self, request, _queryset):
+        return ImportCSVView.as_view()(request, admin=self)
+
+
+def import_csv_data(csv_file, semester) -> int:
```
markBowman: What happens if there is an error on row 25 of 35? Do we end up with 24 proposals added and everything after 24 ignored? I'm not sure if we should try to make it atomic (all or nothing), or able to process as many rows as possible, and somehow let the users know that one row will have to be done by hand.

jnation3406:
I'm worried that their csv is not consistent each semester. I will link you to last semesters example in slack, but its very different from this semesters. So I don't know if this solution will work for them, but if its self contained and generic then its fine being added

markBowman (approved):
Thank you so much for doing this. Even as it stands with just the happy path coded, it could  save Emily hours of mind numbing cutting and pasting. She would have to visit each proposal page to tidy things up, but I think that's acceptable.
I'm approving, but better minds than mine should check the code.

observation_portal/proposals/admin.py
Fingel: I think it should be made atomic, and then the error added to the message like the success message is here: https://github.com/observatorycontrolsystem/observation-portal/pull/363/changes#diff-589271d4b1fc26eaa650ce007ec35fce9c9142a2d6b650f5e9bc849657f83b3eR323

Fingel:
> I'm worried that their csv is not consistent each semester. I will link you to last semesters example in slack, but its very different from this semesters. So I don't know if this solution will work for them, but if its self contained and generic then its fine being added

Can we ask them to use a consistent format?

markBowman:
> I'm worried that their csv is not consistent each semester. I will link you to last semesters example in slack, but its very different from this semesters. So I don't know if this solution will work for them, but if its self contained and generic then its fine being added

My impression is they will follow a schema if we supply it - or maybe I'm just a hopeless optimist.

Fingel:
I'm not sure the timescale on which this is needed.  I can probably fix it up a bit tonight - make it atomically fail and wrestle poetry. Otherwise I'm sure @jnation3406 could take it over the finish line if it's ASAP.

markBowman:
> I'm not sure the timescale on which this is needed. I can probably fix it up a bit tonight - make it atomically fail and wrestle poetry. Otherwise I'm sure @jnation3406 could take it over the finish line if it's ASAP.

By the end of the week would be great.

observation_portal/proposals/admin.py
Fingel: Will just pass. They can add the PI later if needed.

Fingel:
The transaction is atomic now, so any failure will result in no proposals created:
Screenshot from 2026-07-06 13-30-03

jnation3406 (approved):
I've enhanced the error reporting and validating and tested locally. I think its good to go now, assuming we update the actual csv files to have our real instrument_types in them

PR #364, opened 2026-07-06
Drop python 3.8 suppoprt

Fingel:


jnation3406 (approved):
Also update Readme line

